PR #19, opened 2016-03-22
Editing the build file for the RHEV install doc.

elladeon:


_build_cfg.yml
```diff
@@ -32,6 +32,18 @@ Topics:
...
 
 ---
+Name: Installation
+Dir: install
+Topics: 
+  - Name: Installing on RHEV
+    Dir: installing_on_rhev
+    Topics: 
+      - Name: Installing the Appliance
+      - File: Installing_Appliance
```
abellotti: Hi Deon, the 2 File entries here shouldn't have the - prefix.  asciibinder fails to build with those.

PR #21, opened 2016-04-12
update defining_policies_and_profiles deleting downstream references

CarlosGlez7:


delaluzparra:
@soconcar

sergio-ocon:
Can you please change CloudForms Management Engine to {product-title} too?

Defining_Policies_and_Profiles/master.adoc
```diff
@@ -0,0 +1,16 @@
+= Defining Policies and Profiles
+CloudForms Documentation Team <{product-title}-docs@redhat.com>
```
Fryguy: this doesn't seem right as it's an email address.  Also, I don't think the master.adoc files are supposed to come over like this anyway. @elladeon ?

Fryguy:
I think this also needs the associated entries in _build_cfg.yml.  Additionally, the new file names and directories should be all lower case.

CarlosGlez7:
These are the files that have been uploaded and after the last commit they're completed.

1. defining_policies_and_profiles/master.adoc 
2. defining_policies_and_profiles/topics/editing_policy_event_assignments.adoc 
3. defining_policies_and_profiles/topics/to_create_a_condition_to_check_file_contents_of_a_host.adoc
4. defining_policies_and_profiles/topics/to_create_a_compliance_policy.adoc 
5. defining_policies_and_profiles/topics/to_create_an_assign_profile_to_analysis_task_action.adoc
6. defining_policies_and_profiles/topics/to_create_an_snmp_action.adoc 
7. defining_policies_and_profiles/topics/to_create_an_e-mail_action.adoc 

@delaluzparra @sergio-ocon @alexvkcr @katafira

delaluzparra:
@miq-bot add_label documentation

Fryguy:
Still needs :point_up: https://github.com/ManageIQ/manageiq_docs/pull/21#issuecomment-208999642

Also my comment about the master.adoc was collapsed, but I don't think that file is needed at all.  It acts as a table of contents for RH docs team, but here we have the _build_cfg.yml instead.  @elladeon Is that accurate?

Defining_Policies_and_Profiles/master.adoc
elladeon: @Fryguy yeah, we don't need that master.adoc file at all.

elladeon:
@Fryguy, correct, we do not need master.adoc since we're building with asciibinder here. Has there been a commit yet for the _build_cfg.yml file?

defining_policies_and_profiles/master.adoc
```diff
@@ -1,5 +1,5 @@
 = Defining Policies and Profiles
-CloudForms Documentation Team <{product-title}-docs@redhat.com>
+{product-title} Documentation Team <{product-title}-docs@{product-title}.com>
```
elladeon: The team name is right, but that email is wrong. @Fryguy, do we have an entity like {author-email} that we can use? I don't think we can abstract individual segments like that.

defining_policies_and_profiles/topics/editing_policy_event_assignments.adoc
```diff
@@ -8,7 +8,7 @@ The policy evaluates its scopes and conditions when specified events occur in yo
 
 . Click image:../images/1847.png[image] (*Configuration*), image:../images/1880.png[image] (*Edit this Policy's Event assignments*).
 
-. Check all the events you want to assign to this policy. For a description of the events, see link:https://access.redhat.com/documentation/en/red-hat-{product-title}/version-4.1-beta/defining-policies-and-profiles/#chap_events[Events].
+. Check all the events you want to assign to this policy. For a description of the events, see link:https://access.redhat.com/documentation/en/{product-title}-{product-title}/version-{product-version}-beta/defining-policies-and-profiles/#chap_events[Events].
```
elladeon: This is another one where we can't do a simple search / replace. Instead, use conditional text, like 
The documentation is at 
ifdef::manageiq[]
http://whatever
endif[]
ifdef::cfme[]
http://access.redhat.com/blah
endif[]

alexvkcr: Yeah, it makes sense.

What are the links we should make the conditional statements print?

@CarlosGlez7 @katafira @sergio-ocon @delaluzparra

alexvkcr:
Regarding master.doc, ok, we're going to pare it away in next commit in this PR. @Fryguy 

Regarding de _build_cfg.yml file. @elladeon  These new files don't appear there... yet. If requested, we can modify the file according to the rules that appear in the file. But some contribution from others might be requested.

@katafira @CarlosGlez7 @sergio-ocon @delaluzparra

CarlosGlez7:
updated name of PR

CarlosGlez7:
Closing PR, changes go to #50

PR #27, opened 2016-04-20
update integration_with_red_hat_satellite deleting downstream references

katafira:
*************************************************************************************************************
Issue: 
	Do not know if "Red Hat Satelllite 6" shoud be changed. I removed "Red Hat" and left "Satellite 6", as it is specified in the table, but i will need a confirmation if it's ok.

*************************************************************************************************************


katafira:
Closed pr 23. #53 is the updated one.

PR #35, opened 2016-04-26
update release_notes deleting downstream references 

CarlosGlez7:
I had several doubts with the "Red Hat" word. I added a file called "doubts" where you can find the problems that I talk about.


CarlosGlez7:
Closing PR, instead we have #47

PR #45, opened 2016-05-26
[auth] Fixes in the SAML setup instruction documentation

abellotti:
- wrong path specified for Apache template files
- needed the full path for mellon_create_metadata.sh
- needed an Apache restart step


abellotti:
@mfeifer when you get a chance,  minor issues to update as I was re-running through the steps here for the summit.  Thanks.

PR #59, opened 2016-07-18
Added the Appliance Hardening Guide from the CFME repository.

adahms:
Migrating the Appliance Hardening Guide from the downstream repository to test the current setup.


Fryguy:
@adahms So a few comments in on particular order.

- There are a number of references to CFME both in filenames, filepaths, and in the doc itself that I would not have expected, but perhaps there are reasons.
- The first document listed is named Configuring-cloudforms.  I would expect the filename to be product agnostic and be something like "Configuring", and the internally the title can be `== Configuring {product-title}`
- There are files that don't make sense here like `doc-Appliance_Hardening_Guide/cfme/docinfo.xml` and `doc-Appliance_Hardening_Guide/cfme/metadata.ini`
- I don't understand the purpose of files like `common/attributes/cfme.adoc` in the context of ascii_binder except that we don't have these attribute otherwise.  Even so I would expect something like a single file that uses if conditions to pick the right attributes.
- There are no entries in _build_cfg.yml which is the table of contents and dictates how the files are built and TOCs are laid out.
  - Did you run this through AsciiBinder?
  - By extension, files like `doc-Appliance_Hardening_Guide/cfme/master.adoc` are not necessary

common/Configuring_CloudForms.adoc
```diff
@@ -0,0 +1,155 @@
+[[Configuring-cloudforms]]
+== Configuring {product-title}
```
nhr: Top level headers should be L1 (one '=' only):

    = Configuring {product-title}

common/Configuring_CloudForms.adoc
```diff
@@ -0,0 +1,155 @@
...
+
+Although the {product-title} appliance comes configured to be integrated immediately into your environment, you can make some changes to its configuration.
+
+[NOTE]
+======
```
nhr: By convention we limit block delimiters to four characters:

    ====

instead of

    ==============

common/Configuring_CloudForms.adoc
```diff
@@ -0,0 +1,155 @@
...
+The {product-title} appliance is intended to have minimal configuration options.
+======
```
nhr: Same here: re: delimiter length of four characters.

Fryguy:
Also, minor, but general pattern is to keep all filenames underscorized_lower_case and matching the titling of the document...not really required as the _build_cfg.yml handles that, but the convention keeps it from being surprising.

Also note that I have not upgraded to the latest ascii_binder yet, but `_build_cfg.yml` is being renamed to `_topic_map.yml` in the newest version.  I'll keep you in the loop when I upgrade.

common/Configuring_CloudForms.adoc
```diff
@@ -0,0 +1,155 @@
...
+
+=== Changing Configuration Settings
```
nhr: Per the article title comment above, headers within the doc should start at L2 (two '='):

    == Changing Configuration Settings

common/attributes/cfme.adoc
```diff
@@ -0,0 +1,6 @@
+
+:product-title: Red Hat CloudForms
```
nhr: `product-title` is a reserved parameter. AsciiBinder sets this for you based on the `name` parameter in your `_distro_map.yml` file (see this [example](https://github.com/openshift/openshift-docs/blob/master/_distro_map.yml#L3))

common/attributes/miq.adoc
```diff
@@ -0,0 +1,6 @@
+
+:product-title: ManageIQ
```
nhr: See above comment re: reserved `product-title` parameter.

common/Configuring_CloudForms.adoc
nhr: ...by 'top level' here, I mean the article title. Note also that there's no metadata here, but I think you might want to include some - see the first several lines of [this example](https://raw.githubusercontent.com/openshift/openshift-docs/master/architecture/infrastructure_components/kubernetes_infrastructure.adoc).

doc-Appliance_Hardening_Guide/cfme/master.adoc
```diff
@@ -0,0 +1,30 @@
+= Appliance Hardening Guide
+:vernum: 4.1
```
nhr: Use the reserved `{product-version}` parameter here, whose value comes from each branch's `name` value in the `_distro_map.yml` file (see [example](https://github.com/openshift/openshift-docs/blob/master/_distro_map.yml#L36))

doc-Appliance_Hardening_Guide/cfme/master.adoc
```diff
@@ -0,0 +1,30 @@
...
+:doctype: book
+:toc: left
+:toclevels: 3
+:numbered!:
+:experimental:
```
nhr: The rest of these settings are not currently used by AsciiBinder, as the documents are cast into the page template which is located at [_templates/page.html.erb](https://github.com/ManageIQ/manageiq_docs/blob/master/_templates/page.html.erb) inside this docs repo.

doc-Appliance_Hardening_Guide/cfme/master.adoc
```diff
@@ -0,0 +1,30 @@
...
+
+// Custom attributes
+include::common/attributes/cfme.adoc[]
+
+:cfme:
+
+:numbered:
+:leveloffset: 1
+include::topics/Introduction.adoc[]
```
nhr: There's nothing functionally wrong with this, but I just want to note that including a group of topics together in a single document is not consistent with the "topic-oriented documentation" model, which has been adopted by other docs teams in Red Hat.

doc-Appliance_Hardening_Guide/topics/Appliance_Security.adoc
```diff
@@ -0,0 +1,11 @@
+= Appliance Security
+
+:leveloffset: 2
```
nhr: It is not clear to me why you want to offset headers by 2 here. Normally this is only useful when you are nesting sub-topics and their sub-topics together into a skeleton document.

adahms:
@Fryguy @nhr 

Thank you for your feedback.

There are a number of items here that I agree with, such as the number of characters included in delimiters, the use of level offset, file names, and the need to include meta data at the start of the guides. The main focus in preparing content at the moment is to ensure we have content that is available and that we can start to use, but we can work on cleaning this up as well and define some conventions around that as we move forward.

In regards to certain other aspects, we are not using AsciiBinder for the downstream content, and also need to include certain references to CloudForms as a downstream offering when we define that downstream content, such as in the .adoc files in the **common** directory.

To provide some context, the master.adoc file is our main table of contents for a given title, and the docinfo.xml file provides meta data required by the downstream tool chain. I have included these in the upstream version in case they are needed, but we can easily remove the miq version of the docinfo.xml file and entity declarations in the master.adoc without affecting the downstream content. Likewise, we can remove the miq version of the master.adoc file entirely and fill out a _topic_map.yml file if need be for AsciiBinder.

Let me know what you think.

Kind regards,

Andrew

Fryguy:
Regardless, this documentation won't build upstream without the _buid_cfg.yml changes.

Also, not using AsciiBinder downstream completely defeats the purpose of using AsciiBinder at all.  I thought we discussed this already.  cc @chessbyte @geertj

Fryguy:
> and also need to include certain references to CloudForms as a downstream offering when we define that downstream content

@adahms Perhaps there's some confusion on what AsciiBinder provides?  Its purpose is to allow both upstream and downstream documentation to live side by side.  There will always be CloudForms specifics, but those can be handled with the AsciiBinder conventions.  @nhr Can you give some links to some of the OpenShift docs that exemplify this, as an example?

nhr:
Sure. Here's an example of how we use `ifdef` statements in combination with _distros_ that have been defined in `_distro_map.yml` to produce two different versions of the same page:

In the source, we use `ifdef` blocks to add alternate content:

* For OpenShift Enterprise: https://github.com/openshift/openshift-docs/blame/master/install_config/imagestreams_templates.adoc#L16-L25
* For OpenShift Origin: https://github.com/openshift/openshift-docs/blame/master/install_config/imagestreams_templates.adoc#L27-L37

And here are the results when they are generated and published:

* Enterprise: https://docs.openshift.com/enterprise/latest/install_config/imagestreams_templates.html#overview
* Origin: https://docs.openshift.org/latest/install_config/imagestreams_templates.html#overview
 
We use this technique to build and publish four distinct distributions of the documentation on a nightly basis. Once a distribution is defined in the distro map, you can start using ifdefs to provide alternate content as necessary.

chessbyte:
@adahms Do not understand what the connection between upstream documentation in this repo is to downstream CF.  If there is no connection, then we have a problem.

adahms:
@Fryguy @chessbyte 

Thank you for your comments. There are some sections of the adoc files where we do indeed use ifdef statements to define content that is only for CloudForms or only for MIQ, and we do understand the need to use this functionality to conditionalize content for the different brands.

There are a number of topics in here that I think we should discuss in a little more detail, and I would like to start a thread where we can lay everything out and discuss the best approach. There is some good feedback here that I agree we need to address, and I will outline some of the other considerations that I can see from my side as well.

Fryguy:
Closing since this is part of #62

PR #62, opened 2016-07-29
CFME 41 Migration

adahms:
Pull request to migrate the Red Hat CloudForms 4.1 documentation in a format that allows the content to be shared across ManageIQ and Red Hat CloudForms.


Fryguy:
@adahms Great work!  Thanks so much for this!

I'm going to merge as is to get the content in.  

> @adahms wrote in email:
>> While this allows the guides to build correctly, I understand this does not necessarily fill out the table of contents in the side bar the way we might want, so I am happy to revisit this and break the entries up at least into chapters.

Yes, I think this is very important in the next follow up PR because as is, it looks very strange in the left nav.  I think the entire TOC could be written in the _build_cfg.yml with some possible minor restructuring since AsciiBinder only allows 3 levels of nesting. (note that @abellotti [opened an issue](https://github.com/redhataccess/ascii_binder/issues/19) on ascii_binder to request 4th level nesting).

abellotti:
we seem to have lost the v2.2.0 of the API specification with this. In the follow up PR will the REST API section be expanded to include all the changes made to api/  ?  

Also, we lost the Ext Auth doc pages.

PR #77, opened 2016-08-11
Updated .gitignore with global ignores for downstream build directories

lbopf:
This update allows build directories created by tools other than AsciiBinder to be ignored.


dayleparker:
Awesome, thanks Lucy!

PR #78, opened 2016-08-12
Removed an incorrect '5.6' version number from Policies and Profiles …

lbopf:
…Guide (CF only)


adahms:
@lbopf 

Looks good to me; merging.

PR #81, opened 2016-08-15
Bz1348740 Ansible

dayleparker:
Hi @suyogsainkar,  
The content on Ansible Tower has been expanded and reworked a bit since it was published for 4.1 GA. Could you please review and merge the content?  

https://bugzilla.redhat.com/show_bug.cgi?id=1348740 has notes from Pavol Kotvan's reviews as well.  
Many thanks,
Dayle


adahms:
@dayleparker 

Hi Dayle - thanks for the pull request! What do you think about the screen capture that is included in this section? Is that something we can do without?

dayleparker:
@adahms I do think we should keep this screenshot, for the sake of consistency with other adding providers procedures, since it's the first screenshot, and for a visua "I'm in the right place" for the reader. The screenshot in the "Refreshing Providers" procedure, however, I think is not that useful.

dayleparker:
@suyogsainkar, I think this is ready to be merged. Thanks for all your suggestions along the way!

adahms:
@dayleparker Roger that - thanks for the confirmation!

dayleparker:
@adahms Thanks for your comment :) I highly agree we should re-evaluate whether we need all the screenshots we have though -- perhaps as a 4.2 goal?

suyogsainkar:
@dayleparker Thanks for all the updates - looks great; merging now.

PR #88, opened 2016-08-18
Removed unused common topic 'Registering and Updating'

lbopf:
This content exists in the General Configuration guide already, and has been excluded from the Release Notes, so the common file is no longer needed.


adahms:
@lbopf 

Looks good to me; merging.

PR #90, opened 2016-08-26
fixed winrm quickconfig command

dayleparker:
From bug: "In section 1.5.1 the command to Enable WinRM for configuration. is "winrm quickconfig" instead of "winrm quick config". "  

@suyogsainkar, would you mind merging this please?  
Thanks so much!


suyogsainkar:
@dayleparker Looking good; merging.

PR #92, opened 2016-08-29
Changed one instance of 'VmWare' to 'VMware'.

adahms:


abellotti:
LGTM. Thanks!!

I don't have merge privileges here though.  /cc @mfeifer

adahms:
@abellotti 

Thanks for the check - this is a quick and easy one, so I'll merge this if it's no issue with yourself. :)

PR #99, opened 2016-09-07
Added step to specify a zone

dayleparker:
On the cf-tech list, David La Motta noticed we were missing a step to specify the zone when adding an OpenShift provider. No bug was created for this, as Loic requested just the addition of one line:

"Select the appropriate Zone for the provider. By default, the zone is set to default."

I looked into putting some info about why to choose which zone, but we've already covered it well in the Deployment Planning and the Gen Config Guides.


suyogsainkar:
@dayleparker Thanks, Dayle. Looks good to me; merging now.

PR #101, opened 2016-09-08
Updated the correct 4.1 repo name in section Registering and Updating…

suyogsainkar:
… CloudForms


dayleparker:
Looks great, Suyog - thanks for making this change!

PR #105, opened 2016-09-13
Removed leveloffset from REST API Guide

lbopf:
Removed leveloffsets from the REST API Guide. (This also fixes the problem of the downstream guide having no table of contents.)


adahms:
Looks good to me; merging.

PR #109, opened 2016-09-13
Removed leveloffsets from Monitoring, Alerts, and Reporting.

adahms:
Removed all leveloffset declarations from Monitoring, Alerts, and Reporting.


lbopf:
The changes look good to me; merging.

PR #119, opened 2016-09-13
Installing on VMWare - fixed delimiters

dayleparker:


adahms:
Looks good to me; merging.

PR #120, opened 2016-09-13
Fixed delimiters in Installing on RHEV guide

dayleparker:


adahms:
Looks good to me; merging.

PR #136, opened 2016-09-27
[EUWE] Example: Add OpenStack provider via API

psachin:
W.r.t
http://talk.manageiq.org/t/how-to-create-openstack-provider-via-api/1688/,
there is no example to add OpenStack provider. Although the request is
similar but response changes with provider. I have also added link to
"Provider Types" so that it is easy to refer the list while making API
request.

Subsequent examples may follow as I come across.

### DIFF
--- a/api/examples/create_provider.adoc
+++ b/api/examples/create_provider.adoc
@@ -1,6 +1,11 @@
 
 [[create-provider]]
-=== Create Provider
+== Create Provider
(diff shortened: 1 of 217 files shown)


suyogsainkar:
Sorry for the delay in reviewing this PR. Thanks for adding these example - looking good to me. Merging.

PR #137, opened 2016-09-27
Show info about cryptic error

sergio-ocon:
When adding Google Cloud Engine with an appliance that is not properly time synchronized (i.e. Docker or Vagrant after a pause/start), you can have difficulties adding the provider.
There is no easy information to be found about the error in the Internet or the documentation


suyogsainkar:
@sergio-ocon Hi Sergio, thanks for adding the note about the error.

suyogsainkar:
@adahms Hi Andrew, all changes look good to me; merging now.

PR #143, opened 2016-10-07
SAML config instructions added

dayleparker:
I've added some docs on SAML authentication to the General Configuration guide, based on the upstream content. (Thanks to @jvlcek @abellotti @amavinag  for your help on this!)

https://bugzilla.redhat.com/show_bug.cgi?id=1364889

Andrew, it looks like there's merge conflicts -- would you mind having a look and merging this when you have a chance? SAML was only added in 4.1, so we will only need to cherry-pick to euwe afterward.  

Many thanks,  
Dayle


dayleparker:
I set this PR up the wrong way - apologies. Deleting.

PR #151, opened 2016-10-21
Bz1382232 snapshots

dayleparker:
Hi Suyog,

I've added a section "3.35. Managing Virtual Machine Snapshots (Red Hat Virtualization Virtual Machines Only)" to the docs on the VM snapshot feature for https://bugzilla.redhat.com/show_bug.cgi?id=1382232 -- can you please review and provide feedback, or merge if it looks OK?

Thank you,
Dayle


dayleparker:
I configured this PR incorrectly - disregard!

PR #152, opened 2016-10-21
Added procedures and screenshot for reconfiguring RHEV VMs

dayleparker:
Hi Suyog,

I've added a section 3.31. Reconfiguring Virtual Machines (VMware and Red Hat Virtualization Virtual Machines Only) to the docs on the VM reconfiguration feature for https://bugzilla.redhat.com/show_bug.cgi?id=1382234 -- can you please review and provide feedback, or merge if it looks OK?

** This is the new PR after some git troubles. **

Thank you,

Dayle


suyogsainkar:
@dayleparker The changes are looking good to me - merging now.

doc-Managing_Infrastructure_and_Inventory/topics/Reconfiguring_VMs.adoc
```diff
@@ -0,0 +1,18 @@
+[[Reconfiguring_VMs]]
+= Reconfiguring Virtual Machines (VMware and Red Hat Virtualization Virtual Machines Only)
+
+Memory, processors, and disks can be reconfigured on existing VMware and Red Hat Virtualization virtual machines using the *Reconfigure this VM* button.
```
masayag: In RHEVM there are few restrictions for supporting 'add/remove disks':
* Supported by RHEVM 4.0 and above
* Only a single bootable disk is supported (limitation on RHEVM side for disks)
* VM should have at list one existing disk to allow adding additional disks (this limitation is due to the fact that we don't allow the user to specify the destination storage on that dialog and we reuse the storage which is associated with the existing disk)

Should the restrictions be mentioned in the doc ?

PR #153, opened 2016-10-21
RHEV VM snapshots procedures added

dayleparker:
Hi Suyog,

I've added a section "3.35. Managing Virtual Machine Snapshots (Red Hat Virtualization Virtual Machines Only)" to the docs on the VM snapshot feature for https://bugzilla.redhat.com/show_bug.cgi?id=1382232 -- can you please review and provide feedback, or merge if it looks OK?

** Note - this is the correct PR to merge after git troubles**

Thank you,
Dayle


suyogsainkar:
@dayleparker The updates are looking good to me; merging to euwe.

PR #154, opened 2016-10-24
Created a note on high availability

dayleparker:
Hi Suyog,

I've added a note on the ability to configure highly available CloudForms, with a link to a reference architecture the engineering team was basing the feature on (in an ifdef for CF).

Please let me know you think this is sufficient, or if you have any comments.  
Many thanks for reviewing!

Cheers,
Dayle


suyogsainkar:
@dayleparker Looking good to me; merging now.

PR #165, opened 2016-11-07
Bz1383157 topology

dayleparker:
Hi Suyog,  

Would you mind reviewing the section, and letting me know if anything needs fixing or clarifying? Any feedback welcome.  

https://bugzilla.redhat.com/show_bug.cgi?id=1383157

http://file.bne.redhat.com/~dayparke/CloudForms/Topology/build/tmp/en-US/html-single/#network_topology

Many thanks,  
Dayle


suyogsainkar:
@dayleparker Hey Dayle, the section reads well and is easy to follow - nice work!

I have only one minor comment on: "Alternatively, on the provider summary page, click Topology under the Overview list." Since we use "list" for dropdown lists, I think we should come up with a different term. What do you think? I would make reference to it as "in the Overview table" or simply "under *Overview*", however, I think it's worth discussing as a team and adding to our UI conventions as we do make references to the entities on the summary page quite frequently in our docs.

dayleparker:
@suyogsainkar Hey Suyog,  

Thanks for the feedback -- I agree totally. I'm in favour of not naming parts of our UI if we don't need to, so, I prefer "Overview". I'll keep an eye out for this in the future, as I'm bound to do it again! :)  

Re-reading the sentence, I realized it wasn't entirely clear what the point of that statement was, so I've edited it as: "Alternatively, you can open the topology widget from the provider summary page by clicking *Topology* under *Overview*." Let me know if you think it needs more tweaking, or if we're good to merge. Thanks so much!

suyogsainkar:
@dayleparker No problem! Thanks so much for the edits - that's perfect :) Merging now.

PR #179, opened 2016-11-23
Bz1383151 ssui updates

suyogsainkar:


AllenBW:
Power state, verbiage and ss, should those be pulled as per https://github.com/ManageIQ/manageiq-ui-service/pull/371 ?

cbudz:
Hi Suyog,

I've reviewed the changes to the SSUI guide and the only recommendations I have for improving flow and readability would be:

Ch 4: Dashboard Tab

- The second sentence in the Note included is a bit unclear about how and why additional data comes into play on the first of each month? Does data roll over from month-to-month? 

5.1.1 Accessing the Cockpit interface

Here I think its a little difficult to identify the cockpit button on the screen capture, so maybe consider removing the image and the sentence that appears after the Note, and simply starting the instruction? I think using the button image is just as helpful in directing users to it. 

7 Service Catalog:

The last sentence in the guide could probably be broken into two sentences. Maybe the segment beginning with "but you want to add it anyway" could be its own brief instruction like, "Click to add the duplicate item if necessary". 

Other than those suggestions I think it looks great. Let me know what you think.

-Chris

PR #183, opened 2016-12-05
Updated the section title and added info on centralized admin configu…

suyogsainkar:
…ration


doc-General_Configuration/topics/Configuration.adoc
```diff
@@ -2797,9 +2799,10 @@ The region number must be unique on each {product-title} instance where replicat
 [[configuring-a-remote-copy]]
 ===== Configuring a Remote Copy
 
-Configure a {product-title} instance to act as a remote copy where data from the global copy will be replicated. This procedure must only be performed on instances where the *Database Synchronization* role will be enabled. This procedure must be performed before enabling that role.
+Configure a {product-title} instance to act as a remote copy where data from the global copy will be replicated.
```
carbonin: This is a bit ambiguous. It almost makes it seem like the data is coming *from* the global region.

Maybe:
> Configure a {product-title} instance to act as a remote copy from which data will be replicated to the global copy.

doc-General_Configuration/topics/Configuration.adoc
```diff
@@ -2809,22 +2812,27 @@ Configure a {product-title} instance to act as a remote copy where data from the
 [[configuring-the-global-copy]]
 ===== Configuring the Global Copy
 
-Configure a {product-title} instance to act as the global copy from which data is replicated to remote copies. This procedure must only be performed on servers where the *Database Synchronization* role will be enabled. These settings must be completed before enabling that role.
+Configure a {product-title} instance to act as the global copy from which data is replicated to remote copies.
```
carbonin: Similar to above. Maybe:

> Configure a {product-title} instance to act as the global copy to which data is replicated from the remote copies.

doc-General_Configuration/topics/Configuration.adoc
```diff
@@ -2809,22 +2812,27 @@ Configure a {product-title} instance to act as a remote copy where data from the
...
 
-. Navigate to menu:Settings[Configuration].
+image:centralized_admin_enable.png[Enable Centralized Administration]
```
carbonin: Maybe this image would be better if it showed the subscription menu expanded so that the enable option is visible.

doc-General_Configuration/topics/Configuration.adoc
```diff
@@ -2809,22 +2812,27 @@ Configure a {product-title} instance to act as a remote copy where data from the
...
 . Select *Global* from the *Type* list.
-. Add a subscription:
-.. Click *Add Subscrption*.
+. Click *Add Subscription*.
 .. Enter the name of the database on the remote copy in the *Database* field.
 .. Enter the IP address or fully qualified domain name of the remote copy in the *Host* field.
 .. Enter the name of a database user able to access the database in the *Username* field.
 .. Enter the password of the database user in the *Password* field.
 .. Enter the port by which the database is accessed in the *Port* field.
+.. In *Central Admin Enabled*, select *Yes* to enable central administration.
```
carbonin: This is not correct. Central Administration can only be enabled after the initial data sync is completed. So the subscription has to be saved before this can be done.

Also, this is not enabled through the switch on the UI, but from the menu on the right of the subscription row.

doc-General_Configuration/topics/Configuration.adoc
suyogsainkar: Modified; thanks.

doc-General_Configuration/topics/Configuration.adoc
suyogsainkar: Modified; thanks.

doc-General_Configuration/topics/Configuration.adoc
```diff
@@ -2809,22 +2812,29 @@ Configure a {product-title} instance to act as a remote copy where data from the
...
+image:centralized_admin_disabled.png[Centralized Administration Config]
 .. Enter the name of the database on the remote copy in the *Database* field.
 .. Enter the IP address or fully qualified domain name of the remote copy in the *Host* field.
 .. Enter the name of a database user able to access the database in the *Username* field.
 .. Enter the password of the database user in the *Password* field.
 .. Enter the port by which the database is accessed in the *Port* field.
+.. In *Actions*, click the menu on the right to see a list of actions, including *Enable Central Admin*, *Validate*, and *Delete*. You can also *Update* your existing subscription.
+* Click *Enable Central Admin*. Note that central administration can only be enabled after the initial data sync is complete. So you must save the subscription before this can be done. 
+* If *Central Admin Enabled* shows _Yes_, that is if centralized administration is currently enabled, you can *Disable Central Admin* or *Re-Enable Central Admin*. You can also *Validate* or *Delete* your subscription.
+image:centralized_admin_enabled.png[Centralized Administration Enabled]
```
carbonin: This is confusing. You say that this has to be done after clicking save but the instruction to click save is after this in the list. Maybe move these instructions into a section after adding the subscription?

doc-General_Configuration/topics/Configuration.adoc
```diff
@@ -2809,22 +2812,29 @@ Configure a {product-title} instance to act as a remote copy where data from the
...
 . Click *Save*.
 
+Database replication and centralized administration is now configured on your instances. Centralized administration supports life cycle management operations, including virtual machine power operations and retirement, that can be initiated from the global copy and processed and executed on the remote copy. See the section on _Centralized Administration_ in the https://access.redhat.com/documentation/en/red-hat-cloudforms/4.2/single/deployment-planning-guide/[Deployment Planning Guide] for an overview of the feature.
```
carbonin: >Database replication and centralized administration **are** now configured ...

doc-General_Configuration/topics/Configuration.adoc
suyogsainkar: @carbonin Thanks for the feedback. When I clicked the subscription menu, it did not easily expand to show the actions (on my Chrome browser at least). I am now able to access it after a few attempts. I've updated the steps and screen capture accordingly. Please review - thanks.

doc-General_Configuration/topics/Configuration.adoc
suyogsainkar: @carbonin I agree, thanks. I've made a few edits to the existing section. Also added a separate section about enabling central admin after adding the subscription. Let me know what you think. Thanks.

doc-General_Configuration/topics/Configuration.adoc
suyogsainkar: Fixed.

dayleparker:
Looks great Suyog - merging. :)

PR #185, opened 2016-12-09
Bz1390862

cbudz:
Hi Dayle,

I've added some text here to the server roles about enabling SmartProxy Affinity for zones under 719. I'm not entirely sure why this file is showing so many previous changes I didn't make to the file. If you have any insight into this you could share I'd greatly appreciate it. 

Thanks,
Chris


dayleparker:
Hi Chris,
The addition looks good! I left a comment asking for a little more information/specifics -- let me know if you agree/disagree. But at the same time, I don't actually know too much about the SmartProxy and SmartState Analysis and how they entirely relate. :( 

I may have an idea about why your commit is showing so many extra changes -- it looks like your text editor is removing extra white spaces at the end of lines. Do you use Atom, by chance? There's a default package in the program called "whitespace" that is best to disable, for our purposes. See if that works and let me know. 
Cheers,
Dayle

dayleparker:
Looks great -- thanks so much, Chris!

PR #98, opened 2016-09-05
Added 'providing feedback' paragraph to abstract of all docs

dayleparker:
From https://bugzilla.redhat.com/show_bug.cgi?id=1371753:  

Added a shortened paragraph to the abstract of all docs on providing feedback. Where there was already instructions on submitting docs feedback, removed it (Deployment Planning Guide).  

Suyog, thank you for your initial advice with wording, and for reviewing this PR.  :)

PR #191, opened 2016-12-21
Bz1382237

cbudz:
Hi Dayle,.

Please find included here the Installation guide for AWS. I've created a preview here:

http://file.rdu.redhat.com/~cbudzilo/CloudForms/awsGuide/build/tmp/en-US/html-single/

Thanks,
Chris


dayleparker:
Hey Chris,

The doc looks great and looks to be ready to merge. Nice work!

I noticed two small things which aren't a dealbreaker, but might be good to look at-
- should we add command prompts to the commands in 1.3? (ie do you have to be a root user, or an ordinary user) So # or $ ?
- In step 3, should REPONSE be RESPONSE?

There is also an OpenStack chunk of content attached to this (causing merge conflicts). Do you want to create a new PR just for this guide alone, and submit the OpenStack content with something else?

Let me know how you want to proceed with this one.
Cheers,
Dayle

cbudz:
Hi Dayle,

Thanks for reading through the guide. Regarding the command prompt, this is where the IAM role comes in and enables users to run commands on the [S3 bucket](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html). Its specified as a prerequisite, but do you think I should add a nothing command, like "ensure you have created the proper IAM role before continuing?"

The second instance was a typo which I've corrected. 

The new PR for the guide, sans the unwanted matter can be found at PR #[198](https://github.com/ManageIQ/manageiq_docs/pull/198)

Thanks for your help on this.
Chris

dayleparker:
Hey Chris,
Re: the roles - thanks for the explanation! I actually don't know much about working with AWS at all.
I'm not sure if it would be obvious to an AWS person to be using that role... probably wouldn't hurt to add a note about ensuring you're using that role, but it also reads fine as it is.
I'll merge it as it is now in #198, but if you want to add a note later and make a new, feel free to assign it to me and I'll merge that one after the holidays :)
Thanks for your work on this guide!
Cheers,
Dayle

dayleparker:
Closing as the new PR is in #98.

PR #192, opened 2016-12-21
Bz#1383132

cbudz:
Hi Dayle,

Please see the newly created instructions for adding nodes in the OpenStack environment per [BZ1383132](https://bugzilla.redhat.com/show_bug.cgi?id=1383132). 

 I've created a preview here:

file:///home/cbudzilo/manageiq_docs/doc-Managing_Providers/cfme/build/tmp/en-US/html-single/index.html#adding_openstack_cloud_providers

Do you think this should be its own topic, instead of nested under the Adding OpenStack Provider?

Thanks,
Chris


dayleparker:
Hey Chris,
It looks great. And since the reigstering nodes seems like a thing you'd do right after adding an openstack provider, i think it works here. Let's merge it like it is, and if we want to change it in the future, that's cool too.
Cheers,
Dayle

PR #198, opened 2016-12-21
Bz1382237 updated

cbudz:
Hi Dayle,

Here's the new PR with the AWS installation guide, with the suggested changes.

-Chris


dayleparker:
Looks great, thanks so much, Chris!

PR #200, opened 2017-01-12
Removed all references to the Database Synchronization role and the r…

suyogsainkar:
…eplication worker that no longer exist.

PR #202, opened 2017-01-16
Removed all references to the database synchronization role and repli…

suyogsainkar:
…cation worker as they no longer apply.

Note: This PR is for CF 4.1 (Darga) version of the the guide. PR for CF 4.2 (Euwe): https://github.com/ManageIQ/manageiq_docs/pull/200


carbonin:
The replication worker was only removed in 4.2, it was deprecated in 4.1, but was still present and usable.

I'm not sure how you want to handle that kind of situation from a docs perspective.

suyogsainkar:
@carbonin I've added a note at the beginning of the section (https://github.com/ManageIQ/manageiq_docs/pull/205) and in the release notes.

PR #210, opened 2017-01-25
Bz1411945 dedicated db

dayleparker:
@carbonin
Hi Nick,

Thank you for the info about the appliance console configuring an internal DB in https://bugzilla.redhat.com/show_bug.cgi?id=1411945. I've applied some edits to incorporate your info; would you mind reviewing this procedure and letting me know if it needs corrections?

Cheers,
Dayle

### DIFF
--- a/common/configuration-db.adoc
+++ b/common/configuration-db.adoc
@@ -12,16 +12,19 @@ Before installing an internal database, add a disk to the infrastructure hosting
 * If this is the first {product-title} appliance, choose *1) Create key*.
 * If this is not the first {product-title} appliance, choose *2) Fetch key* from remote machine to fetch the key from the first {product-title} appliance. All {product-title} appliances in a multi-region deployment must use the same key.
 . Choose *1) Internal* for the database location.
-. Choose a disk for the database. For example:
+. Choose a disk for the database. This can be either a disk you attached previously, or a partition on the current disk. 
+For example:
 +
 ----
-1)  /dev/vdb: 20480
-
-Choose disk:
+1) /dev/vdb: 20480
+2) Don't partition the disk 
 ----
 +
-Enter *1* to choose `/dev/vdb` for the database location.
-
+* Enter *1* to choose `/dev/vdb` for the database location. This selects the default disk shown to use for the database.
+* If you choose *Don't partition the disk*, a prompt will confirm this choice. (Why does the menu offer choice #2 to not partition the database disk?  Why is there a menu choice to *not* partition it?  What happens if I select this choice?)
+. Enter *Y* or *N* for *Configure this server as a dedicated database instance?*
+* Select *Y* to configure the appliance only as a database. The resulting appliance will not have a user interface. This option is the default, and also can be chosen by pressing *Enter*.
+* Select *N* to configure the appliance with a user interface.
 . When prompted, enter a unique three digit region ID to create a new region.
 +
 [IMPORTANT]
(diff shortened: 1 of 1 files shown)


common/configuration-db.adoc
```diff
@@ -12,23 +12,32 @@ Before installing an internal database, add a disk to the infrastructure hosting
...
+2) Don't partition the disk 
 ----
 +
-Enter *1* to choose `/dev/vdb` for the database location.
-
-. When prompted, enter a unique three digit region ID to create a new region.
+* Enter *1* to choose `/dev/vdb` for the database location. This option creates a logical volume using this device and mounts the volume to the appliance in a location appropriate for storing the database. The default location is `/var/opt/rh/rh-postgresql95/lib/pgsql`, which can be found in the environment variable `$APPLIANCE_PG_MOUNT_POINT`.
+* Enter *2* to continue without partitioning the disk. A second prompt will confirm this choice. Selecting this option results in using the root filesystem for the data directory (not advised in most cases). 
+. Enter *Y* or *N* for *Configure this server as a dedicated database instance?*
+* Select *Y* to configure the appliance _only_ as a database. The resulting appliance will not have a user interface. This option is the default, and also can be chosen by pressing *Enter*.
```
carbonin: I don't think there is a default at this point.

In the next errata release a default is added (N) and more detail is added to the prompt so I think this will need to change.

The new prompt can be seen in https://github.com/ManageIQ/manageiq-gems-pending/pull/27

common/configuration-db.adoc
```diff
@@ -12,23 +12,32 @@ Before installing an internal database, add a disk to the infrastructure hosting
...
+* Select *N* to configure the appliance with a user interface.
```
carbonin: I'm not sure what the best wording would be here. Do you think "user interface" as the differentiating factor is the most accessible for users?

I'm just concerned that they might think this means that the application is started, but without the "User Interface" role. In reality the application is not started (nor configured) on this server. This essentially gives you a basic PostgreSQL server, as if manageiq was not installed at all.

common/configuration-db.adoc
dayleparker: The new prompt looks good! I've removed the extra info about a default and hitting Enter now.

common/configuration-db.adoc
dayleparker: @carbonin I see what you're saying. We refer to the graphical CloudForms accessed through a web browser as "the user interface" (lower case) in the docs, so I would think it would be understandable to readers. However, I was unaware of what was actually happening in the background. I learned that in the SSUI guide, we occasionally differentiate UI/SSUI with "the full administrative user interface". What do you think of:

* Select *Y* to configure the appliance _only_ as a database. As a result, the appliance is configured as a basic PostgreSQL server, without a user interface. 
* Select *N* to configure the appliance with the full administrative user interface.

When you say "the application", are you referring to the appliance?

dayleparker:
@carbonin sorry for the delay on this -- I've made a few small changes; please let me know your thoughts.

carbonin:
Looks good @dayleparker :+1:

dayleparker:
Awesome, thanks @carbonin!

dayleparker:
Hi Suyog,

Would you mind reviewing and merging if this change looks OK to you?

Many thanks,
Dayle

suyogsainkar:
@dayleparker Hi Dayle, looking good to me; merging now.

PR #213, opened 2017-01-31
Installing on X structural fixes

dayleparker:
As I was republishing all the "Installing on X" books, I found that the GCE and Azure guides did not source the configuration chapter from the common folder.

I've updated these links for both guides, in miq and cfme to link to common configuration.adoc topic.

* Without making this change, the guide looks like:
https://doc-stage.usersys.redhat.com/documentation/en/red-hat-cloudforms/4.2/single/installing-red-hat-cloudforms-on-google-compute-engine/

I also found that there was a VMware VDDK chapter in the Openstack guide, duplicated from the "Installing on vSphere" -- so I've commented it out for now on the slim chance it might still be needed.

The previous version of the RHOS guide looks like this (see chapter 3):
https://doc-stage.usersys.redhat.com/documentation/en/red-hat-cloudforms/4.2/single/installing-red-hat-cloudforms-on-red-hat-enterprise-linux-openstack-platform/

Suyog, apologies for the second PR for this bug -- would you mind approving if you're ok with the changes?

Thank you  so much! :)

### DIFF
--- a/doc-Installing_on_Google_Compute_Engine/cfme/master.adoc
+++ b/doc-Installing_on_Google_Compute_Engine/cfme/master.adoc
@@ -18,7 +18,7 @@ include::common/attributes/cfme.adoc[]
 include::topics/installation.adoc[]
 
 [[Configuration]]
-include::topics/configuration.adoc[]
+include::common/configuration.adoc[]
 
 [[initial-login]]
 == Logging In After Installing {product-title}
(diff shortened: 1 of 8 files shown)


dayleparker:
OK, new comment - I've chatted with Suyog and there was a reason for the configuration chapter to be different (to be investigated).

I've reversed those changes and have now only updated the procedure in Configuring an Internal Database in the GCE and Azure guides. (everything else will be investigated in https://bugzilla.redhat.com/show_bug.cgi?id=1417797)

suyogsainkar:
@dayleparker Thanks, Dayle. Looking good to me; merging now.

dayleparker:
@suyogsainkar  Thanks so much, Suyog!

PR #219, opened 2017-02-06
`cf-hap1` was wrongly named to `cf-hap2`

psachin:


suyogsainkar:
@psachin Hi Sachin, thanks for the pull request. The change looks good to me; merging to euwe now.

PR #220, opened 2017-02-07
Bz1379977 appliance console

suyogsainkar:


common/appe-appliance-console-cli.adoc
```diff
@@ -0,0 +1,410 @@
...
+							Description
+
+|
+
+							--region (-r)
+
+
+|
+
+							create a new region in the database (requires database credentials passed)
```
carbonin: This feels like it should more clearly state that the region number is given as an argument for this option.

suyogsainkar: Fixed, thanks.

cbudz:
Hi Suyog,

Everything looks good here. Merging the content.

PR #227, opened 2017-02-16
Removed event VM Discovery as it was removed from CFME since version …

suyogsainkar:
…5.6.


dayleparker:
Looks good to me! Verified https://bugzilla.redhat.com/show_bug.cgi?id=1421959.

PR #228, opened 2017-02-16
Removed VM Discovery from the Events and Description table as it's be…

suyogsainkar:
…en removed as an event type from CFME 5.6


dayleparker:
Looks good to me! Verified https://bugzilla.redhat.com/show_bug.cgi?id=1421959.

PR #237, opened 2017-02-23
[EUWE] Remove duplicated filenames (only differ in case)

Fryguy:
@adahms Please review.  Note this PR is direct to euwe as it only occurs on that branch.

Both of these files have duplicated filenames that only differ in case.  I'm not sure how they got that way, but I am removing the unused ones.  I grepped the code and found the ones that are *actually* referenced, and removed the other one.

This causes me a ton of grief when I am tagging releases because when I checkout a different branch, such as darga, and then switch to euwe, git loses its mind because it won't checkout both files.  My release script is then left in a bad state.

Here are the files at the point of the PR, so you can see the duplicates: 
https://github.com/ManageIQ/manageiq_docs/tree/9c5d52d558925c08b1ddae25abe14d8f01311a09/doc-Introduction_to_Self_Service_Portal/images


adahms:
@Fryguy - Thanks for the patch!

Looks good to me; merging.

PR #245, opened 2017-03-01
replace all instances of deprecated name

cbudz:
Have removed all instances of Red Hat Access Insights and replaced with Red Hat Insights.

### DIFF
--- a/doc-General_Configuration/topics/RH_Access_Insights.adoc
+++ b/doc-General_Configuration/topics/RH_Access_Insights.adoc
@@ -1,31 +1,31 @@
 
 [[rh-access-insights]]
-== Red Hat Access Insights
+== Red Hat Insights
 
-*Red Hat Access Insights* is a new service that uses the collective knowledge to help end users proactively diagnose systems and avoid critical downtime situations. Red Hat Access Insights does this by having systems periodically check in similar to Red Hat Subscription Management.
+*Red Hat Insights* is a new service that uses the collective knowledge to help end users proactively diagnose systems and avoid critical downtime situations. Red Hat Insights does this by having systems periodically check in similar to Red Hat Subscription Management.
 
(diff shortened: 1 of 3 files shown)


doc-General_Configuration/topics/RH_Access_Insights.adoc
```diff
@@ -1,31 +1,31 @@
 
 [[rh-access-insights]]
-== Red Hat Access Insights
+== Red Hat Insights
 
-*Red Hat Access Insights* is a new service that uses the collective knowledge to help end users proactively diagnose systems and avoid critical downtime situations. Red Hat Access Insights does this by having systems periodically check in similar to Red Hat Subscription Management.
+*Red Hat Insights* is a new service that uses the collective knowledge to help end users proactively diagnose systems and avoid critical downtime situations. Red Hat Insights does this by having systems periodically check in similar to Red Hat Subscription Management.
```
dayleparker: @cbudz Do you think Red Had Insights is old enough now that we should remove the word "new"? Might also future-proof our docs a bit as well.

dayleparker:
@cbudz Hey Chris,
The changes look good to me and I did a quick ctrl+F to check there were no lingering instances of "Access Insights". Only thing I'm wondering if we should also fix/investigate while we're in here anyway, is calling it a "new" feature, and whether it's still tech preview. (I added a comment about this). What do you think?
Cheers,
Dayle

cbudz:
Hey Dayle,

I'm all in favor of whatever can reduce future work, so I'll look into your recommendations and update based on what feedback I get.

-Chris

cbudz:
Hey Dayle,

I got some feedback from Rob Williams who initially reported this BZ. He said this should remain as tech preview for the time being. 
I did remove the reference to "new" though since its been around for a bit now. 

-Chris

dayleparker:
Hey Chris,
Thanks so much for looking into it -- and good call on removing the 'new'. 
Merging!
Cheers,
Dayle

PR #226, opened 2017-02-15
Bz1414511

cbudz:
Hi Suyog,

Please find here my update to the Deployment Planning guide for database sizing to accommodate for C&U metrics gathering. Let me know if you have any suggestions, especially relating to the table I've included. 

Thanks,
Chris

PR #246, opened 2017-03-01
revise intro doc with reviewer suggestions

cbudz:
Hey Suyog,

I got tied up in a merge conflict and couldn't easily resolve it, so I've just created this new PR for BZ#1414511. I did resolve the preview which you can see here:

http://file.rdu.redhat.com/~cbudzilo/CloudForms/database-reqs/build/tmp/en-US/html-single/#database-requirements

I also looked up 'rollup' and it appears IBM uses rollup metric in the noun form and roll up for verb.


suyogsainkar:
@cbudz Hi Chris, many thanks for making those changes from https://github.com/ManageIQ/manageiq_docs/pull/226 -- looking good to me. 

By the way, now that I have a preview of the overall doc, I have a couple of additional comments, mainly about the newly added table:
1. I wonder if we should increase the text size in the table (image) by one or two and complete the cell borders? I realise there is a lot to fit into those columns but I think it would provide a clearer view of the table contents.
2. Since the section provides guidelines not for any specific environment, do you think we should include the table as an example of the number of records created for a single OpenStack instance - under **Example:**?
3. "Use the following information ~~below~~ to plan for your increased VMDB needs when working with metrics gathering:"

Please let me know what do you think.

Thanks,
Suyog

cbudz:
Hey Suyog,

I've updated here implementing your suggestions. I removed the image and instead replaced with a table, which I think makes it far easier to read.

suyogsainkar:
@cbudz Hey Chris, thanks for making those changes; looking great! Merging to Euwe.

PR #288, opened 2017-03-21
Remove steps from create service dialog

cbudz:
removed steps that throw error when creating service dialog.


suyogsainkar:
@cbudz Hi Chris, thanks for the revised PR in GitHub.The changes look good to me; merging to darga now.

PR #289, opened 2017-03-21
Typo emserverd vs evmserverd

isimluk:
I haven't heard about emserverd and I cannot find it.


adahms:
Assigning @dayleparker to review.

dayleparker:
Good catch -- thanks for finding and fixing these typos @isimluk!

PR #291, opened 2017-03-22
Re-created the symlink to the topics directory for the HA guide.

adahms:


dayleparker:
Thank you for making this fix, Andrew (I was so puzzled).

PR #295, opened 2017-03-27
Updated the General Configuratuion guide in accordance with feedback from l10n.

adahms:


dayleparker:
Looks good to me; merging.

PR #327, opened 2017-04-20
Edits to the Integration with OpenShift guide

bfallonf:
cc @adahms 

What I did:

* Added a "Prerequisites" file
* Left instances of "OpenShift Enterprise" as that instead of "OpenShift Container Platform" in case there's thoughts on doing a batch change later.
* Arranged the Storage information to flow better, but that meant changing some sections around (but it should be ok).

Andrew, please let me know if there's anything else needed here, and I'll make the required changes.

Thanks!

### DIFF
--- a/common/provider-ose-mgt-token-3_1.adoc
+++ b/common/provider-ose-mgt-token-3_1.adoc
@@ -1,19 +1,18 @@
 Run the following to obtain the token needed to add an OpenShift Enterprise 3.1
 provider:
 
-. To obtain the `management` service account token name, run the following command::
+. Obtain the `management` service account token name:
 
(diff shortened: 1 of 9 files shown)


dayleparker:
Hi Brice,
The edits you made look great -- huge thanks for lending our team some help with OpenShift terminology, as well as with general wording!

I updated a few very tiny things, like how we format a pointer to another guide in RHCI, and I also updated a few CloudForms 4.1 docs links to 4.2 (these will need editing, and likely conditionalizing, once again for the current release).

Merging all of these changes now; I'll make a point of following up with @adahms about the last bits:
conditionals, new URLs, and how to proceed with the OCP product name.

Cheers,  
Dayle

PR #332, opened 2017-04-25
Updates to container providers topology feature

cbudz:
Hi Dayle,

This PR includes updates related to [BZ1439973](https://bugzilla.redhat.com/show_bug.cgi?id=1439973) and [BZ1439981](https://bugzilla.redhat.com/show_bug.cgi?id=1439981). Since both involved updates that were related I included them both here, so I hope that's not a problem. Both are rather minor, but please let me know if you think anything requires revision or further explanation.

I've set up a preview of the additions here:

http://file.rdu.redhat.com/~cbudzilo/CloudForms/topology/build/tmp/en-US/html-single/#using-the-topology-widget

Also, I added an openshift provider to our current CF test instance, so you can try out the topology widget for fun. 

Thanks!
Chris


dayleparker:
Hi Chris,
The new content and the re-organization looks great. (And it was fun to try out the topology view on our test instance!)

The only change needed is to make the naming consistent for this menu: Configuration → My Settings. In the General Config guide, we've settled on calling that unnamed menu at the top right the (lower-case) "settings menu" (after some discussion at our meetings a few months back). We might need to add this to our Conventions page if it's not there :/

For example, see 2.1.1 in Gen Config: https://access.redhat.com/documentation/en-us/red_hat_cloudforms/4.2/html-single/general_configuration/

One other idea that occurred to me while reviewing -- would it be worthwhile including a small icon screenshot of the topology button in 7.8.3.2, since the only way to get to to project topology seems to be via that button? (in 7.8.3.2. Viewing the Topology for Container Provider Projects - "click the topology view button. ") Might make it a little easier to find for users unfamiliar with our icons and buttons.

Please let me know your thoughts!
Cheers,
Dayle

cbudz:
Hey Dayle,

I just pushed the updates to the file based on your suggestions. I also added a topology view button image. I've updated the preview, and you can see all changes there. (same link as above).

-Chris

dayleparker:
Hi Chris,
Thanks for making those edits - it looks good with the Topology view button.
Sorry to be a bit unclear with the settings menu comment -- since it's not an actual label in the UI, I should have mentioned it shouldn't be marked up as a GUI label -- I adjusted the markup in the file to save time -- I hope that's OK.
Merging now to master-- thanks for your work on this!
Cheers,
Dayle

PR #334, opened 2017-04-27
Remove the pg_rewind method of reintroducing a cluster server

carbonin:
The following three PRs effectively cover all of this functionality
by removing the existing database on a server and running through
the initial clone process if data exists on a server from a failed
node.

https://github.com/ManageIQ/manageiq-gems-pending/pull/1
https://github.com/ManageIQ/manageiq-gems-pending/pull/126
https://github.com/ManageIQ/manageiq-gems-pending/pull/142

This is much easier and less error-prone than the pg_rewind method.


carbonin:
@miq-bot assign @dayleparker

dayleparker:
Thanks so much for making this docs update, Nick. Looks great :)

PR #342, opened 2017-05-08
add refresh procedure

cbudz:
Hey Dayle,

I've added some steps here about refreshing both an individual playbook repository as well as all repositories in the inventory. You can check out a preview of the additions here:

http://file.rdu.redhat.com/~cbudzilo/CloudForms/refresh/build/tmp/en-US/html-single/#refreshing-repositories

Let me know what you think. 

Thanks!
Chris

### DIFF
--- a/doc-Managing_Providers/topics/Automation_Management_Providers.adoc
+++ b/doc-Managing_Providers/topics/Automation_Management_Providers.adoc
@@ -39,6 +39,22 @@ Add a repository so that {product-title} can discover and make available your Pl
 
 Once you have synced a repository, its Playbooks will become available to {product-title_short}.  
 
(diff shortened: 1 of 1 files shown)


dayleparker:
Hi Chris,
The procedures look good to me and make sense where they are in the guide.
One thought I had while reading -- would it be helpful add the word 'Ansible' to section titles (and the others around it), like "Refreshing Ansible Repositories"? It's not a rule, but I think it could potentially help findability on Google and via other searches. I'd suggest it for all 3 of the subsections within the Ansible section -- but up to you. 
Anyway, I didn't want to hold up this PR for this suggestion, but please let me know what you think :)
Cheers,
Dayle

PR #343, opened 2017-05-08
[EUWE] Bz#1375742 loadbalancer euwe

dayleparker:
Hi Suyog,

I've added a new section to the Deployment Planning Guide on the load balancer config to make it work with the SSUI correctly. Please let me know what you think of the content, if you think it needs any corrections, or you have any suggestions -- feedback welcome! It's not something we've covered before, as far as I could tell... :/

Once this is merged, I'll cherry-pick this over to fine/4.5 as Alberto has confirmed this config is also needed there.
Preview: http://file.bne.redhat.com/~dayparke/CloudForms/Loadbalancer3/build/tmp/en-US/html-single/#load-balancer

Thanks so much!
Dayle

### DIFF
--- a/doc-Deployment_Planning_Guide/topics/planning.adoc
+++ b/doc-Deployment_Planning_Guide/topics/planning.adoc
@@ -672,9 +672,7 @@ standard_conforming_strings = off	# MIQ Value not sure why this is enabled
 
 Deploying several {product-title_short} appliances with a load balancer can be useful for performance.
 
-//Ask SMEs Colin's questions from Comment 9, 10, 11: Any way we can turn off the memcached server running on these hosts? (unrelated resource contention issues suggest this would be useful)
-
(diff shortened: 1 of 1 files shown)


suyogsainkar:
@dayleparker Hi Dayle,

Thanks for adding this new section - the procedure looks good to me. While the content itself looks good, I'm not sure if the Deployment Planning Guide would be the best place for this content; also since engineering are working on a solution that will probably eliminate the need to do this configuration in the next release of CloudForms, so sounds like a solution applicable to certain releases, than a regular deployment planning task. That said, I can't think of a better place myself and had a quick chat with Andrew as well, so let's get it in for now as this is an urgent customer request, and we can always revisit later if required. I'm merged this PR to Euwe.

Also, I have one question and a suggestion as below.

- I think it would be worth bringing the Important box before step 1 which would provide advance notification to customers, and help plan the configuration steps.

- In step 2, "The *memcache_server: option* line, which specifies where it connects to, remains set as itself (127.0.0.1)." I was unable to locate that line in Advanced Configuration. Is it referring to the *memcache_server* or *memcache_server_opts line?

Please let me know what you think, and maybe open a separate PR if required?

Thanks,
Suyog

PR #344, opened 2017-05-08
added two new container roles to groups table

dayleparker:
Hi Chris,

I've added the two new container management roles to the General Config guide table which contains the other roles. Could you please review the changes?

I also fixed a typo ( s/server/service) and noticed we hadn't included Consumption Administrator in the docs yet, so added that role too. I realize we still need a description of this, so have created a new bug as it will require more digging and is a bit out of scope for this bug - https://bugzilla.redhat.com/show_bug.cgi?id=1448784.

If you want to check out the list of roles (I had to dig through our docs to remember how), go to the settings menu > Configuration > Access Control dropdown side menu > expand Groups.

Preview here (ctrl+F section "4.1.4.2.6. Using Groups Named by Red Hat CloudForms to Assign Account Roles"): 
http://file.bne.redhat.com/~dayparke/CloudForms/container-roles/build/tmp/en-US/html-single/

Please let me know if you have any feedback. I checked the Release Notes and luckily we've already mentioned the roles there too.

Cheers,
Dayle

### DIFF
--- a/doc-General_Configuration/topics/Configuration.adoc
+++ b/doc-General_Configuration/topics/Configuration.adoc
@@ -954,17 +954,20 @@ This group must be in the LDAP directory source you specified for the server. Se
 |EvmGroup-administrator|Administrator
 |EvmGroup-approver|Approver
 |EvmGroup-auditor|Auditor
+|EvmGroup-consumption_administrator|Consumption Administrator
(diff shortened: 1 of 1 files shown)


cbudz:
Hey Dayle,

The additions look good. Merging this request. 

-Chris

PR #347, opened 2017-05-09
[EUWE] Moved Important admonition, edited parameter name

dayleparker:
Hi Suyog,

Thanks for your feedback and review, and for merging the original. Could you provide a quick follow-up review, as I've made the changes you suggested.

I've moved that Important box to the beginning as per your suggestion -- please let me know what you think.

RE: "option" in "The *memcache_server: option* line" was just a variable/placeholder for where to specify the IP address -- so this part is referring to the "memcache_server" line, which is just above the "memcache_server_opts" line in the advanced config file. I've removed the word "option" as it might have been confusing.

I will cherry-pick the whole lot of additions/changes to master after.

Preview:
http://file.bne.redhat.com/~dayparke/CloudForms/Loadbalancer4/build/tmp/en-US/html-single/#load-balancer

Many thanks,
Dayle


suyogsainkar:
@dayleparker Perfect - thanks for making those changes, Dayle! Merging to euwe.

dayleparker:
Thanks so much for the second review and your suggestions, Suyog! :)

PR #354, opened 2017-05-12
Added updated "adding a middleware provider" content by Julie

dayleparker:
This PR adds Julie Stickler's revised "Adding a Middleware Provider" section (with SSL options), originally written in CEE Gitlab (internal repo).

I've done a little bit of rejigging to fit it into the existing chapter, removing the "Middleware Management header" and the first step about logging into the appliance, to be consistent with other 'adding a provider' sections.

Tracked by https://bugzilla.redhat.com/show_bug.cgi?id=1437286

### DIFF
--- a/doc-Managing_Providers/topics/Middleware_Providers.adoc
+++ b/doc-Managing_Providers/topics/Middleware_Providers.adoc
@@ -24,9 +24,15 @@ After initial installation and creation of a {product-title} environment, add a 
 . Accept the default *Zone*.
 . Under *Endpoints*, configure the following for the middleware provider:
 .. Select a *Security protocol* method to specify how to authenticate to the provider:
-** *SSL* (SSL with validation) – Authenticate to the provider securely using a trusted certificate authority. This requires that you have already configured your public and private keys in the *_/client-secrets_* directory as either two .PEM files or as a single .pkcs12 file.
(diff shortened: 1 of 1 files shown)


JStickler:
@dayleparker, I just realized, I mentioned the required SSL certificate configuration in this topic, but I should also have added a note that in order to use SSL the MM server must be started with the service option HAWKULAR_USE_SSL=true (I didn't realize it until I saw this topic removed from the rest of my doc).  We should probably add a note.

dayleparker:
Hi @JStickler, 
No problem -- I added a note mentioning just before the SSL options (although at first I had it at the beginning of the section). Is that what you had in mind? Feel free to comment directly on the file in the "Files changed" tab if it's easier too.
I also updated the language of "authenticate to the provider" to "authenticate the provider" -- I spotted it on this round, and Beni had suggested I also change that wording in my SSL updates in https://bugzilla.redhat.com/show_bug.cgi?id=1440602#c12.

Anyway, if you are happy with the content here, please assign a reviewer, or if there are further changes to be made, please let me know.
Cheers,
Dayle

dayleparker:
We discussed in today's docs meeting it's probably fine to merge this one to ensure it gets into the build docs -- and I'll take it on as reviewer since I did a bit of reviewing while adding it to GitHub. 
@JStickler, if you see any additional changes that are needed, just let me know. Thanks for your work on this!

PR #367, opened 2017-05-17
Bz1449906

cbudz:
Hey Dayle,

I've created this PR based on BZ[1449906](https://bugzilla.redhat.com/show_bug.cgi?id=1449906). I added a note under 4.5.2.1. Adding a Service Dialog notfiying users of the *param* prefix requirement when setting variables for Ansible playbook service catalog items. I've created a preview here:

http://file.rdu.redhat.com/~cbudzilo/CloudForms/params/build/tmp/en-US/html-single/#service-dialogs

What do you think of the location? I thought it might be best to note when creating a service dialog, but it could just as easily fall under the main topic. 

-Chris

### DIFF
--- a/common/create-playbook-service-catalog-item.adoc
+++ b/common/create-playbook-service-catalog-item.adoc
@@ -3,7 +3,7 @@ Create a catalog item that uses an Ansible Playbook to back it.
 
 [NOTE]
 ====
-Before creating an Ansible service, at least one repository, one playbook, and one credential must exist in the {product-title} inventory. Check your inventory and add the appropriate resources before creating an Ansible service. For more information on Ansible resounces, please see XXXXX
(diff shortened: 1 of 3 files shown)


dayleparker:
Hi Chris,

Sorry for the delay in my reviewing, The note and its placement looks great to me, reading the topic and where you'd need to know about formatting the variable. (However, I am not super familiar with service dialogs and catalog items... ). We might want to make this an "Important" admonition though, if it means the task doesn't work without this knowledge though, which is the impression I get from the BZ -- what do you think? 

BTW, I spotted a few small issues while reviewing this area of the guide (not part of this PR though):
- The procedure in 5.3.2.1. Adding a Service Dialog isn't building correctly from step 4.
- The note in 5.3.1.1. Creating an Ansible Playbook Service Catalog Item ends with a sentence "For more information on Ansible resounces, please see XXXXX "

Do you want me to create a separate bug for that, or would you mind making those edits if you end up changing the Note to an Important box?

Cheers,
Dayle

cbudz:
Hey Dayle,

No need to make a new bug, I'll just make those changes after updating to to flag the note as important. 

-Chris

cbudz:
Hey Dayle, everything is updated. Changes should be visible in the above linked preview.

dayleparker:
Awesome, thanks for making those edits, Chris! Everything looks great :)

PR #370, opened 2017-05-21
[WIP] Bz#1371751 providerupdate

dayleparker:
This branch adds descriptions of provider types throughout guide and adds a bit of consistency across chapters.

Note when cherry-picking, do not cherry-pick the automation chapter to 4.2.

See https://bugzilla.redhat.com/show_bug.cgi?id=1371751 for details.


dayleparker:
Hi Chris,
Thanks for volunteering to review this one -- I'd love to get your opinions on the definitions and terminology. (And Suyog's as well, I'd forgotten you were originally the QE :S )

I was a bit torn on language while updating these sections -- ie. what's a provider, what's a manager? I'd especially like to hear what you think of the Network Managers description and the Preface. Do you think we should rearrange the book into providers and managers parts? Maybe you have a clearer perspective, having not stared at it as much as I have.

I'm also not sure whether we should edit the subtitle and Abstract of the guide to include 'managers', although mentioning all the managers here would make it a lot more wordy. What do you think?

Also, I had a merge conflict with the automation management provider intro, but resolved it after adding a short intro sentence to make it consistent with the other chapters -- if you want to remove/change it, please go for it, or let me know what I should edit. Would love your thoughts on whether it makes sense or not...

To summarize the changes made for https://bugzilla.redhat.com/show_bug.cgi?id=1371751:

* added a general intro block to the Preface on providers
* added intro paragraphs to any providers which didn't have them - left Storage and Middleware as is as they had intro sentences similar to the other chapters
* added a thumbnail and overview table to Cloud Providers, as that was the only section that had something similar to the one in Infrastructure Providers
* did not add a section on discovering cloud providers as requested in Comment 0, as there were already separate procedures for discovering Azure and Amazon providers.

Preview: file.bne.redhat.com/~dayparke/CloudForms/Providers2/build/tmp/en-US/html-single/

Thanks in advance for your feedback!
Cheers,
Dayle

cbudz:
Hey Dayle,

Here's my feedback on this PR:

- I was a bit torn on language while updating these sections -- ie. what's a provider, what's a manager?etc.

This reads well to me and matches with my understanding of the difference between providers and managers. It does seem to be rather difficult to employ exact language as to the 	differences between providers and managers since they do seem to be very similar or overlap, 	but I think you’re successful in the definitions that you’ve added. I like the OpenStack example 	as a manager, since this page gives a nice and clear summary of how the product manages 	and controls the various resources in the network, vs providers focusing on clouds or  containers, etc.
The only place things get a little confusing (to me) is where we use the language Red Hat Virtualization Manager Provider. Same goes for MicroSoft SCVMM Manager Provider. Maybe this idea of Manger vs Provider is something we can discuss and bounce around with some other groups to lock down concrete language as to how exactly specific products should be referred. I’m not sure entirely, outside of the common providers we encounter, how these system are referred to by people using them.  Taking a quick glace, EC2 System Manager is very specifically referred to as a System Manager. Maybe we can compile a list for internal reference.

Do we need to separate out providers from managers? My assumption is that the organization of the guide into the topics like cloud, infrastructure, network, etc provides the high level structure that users might find most helpful, and its also easily searchable. 

- I'm also not sure whether we should edit the subtitle and Abstract of the guide to include 'managers', although mentioning all the managers here would make it a lot more wordy. What do you think?

I think it might help to just reference providers and system managers in the subtitle but enough information is provided in the preface. I wouldn't mention all the managers. 

- Also, I had a merge conflict with the automation management provider intro
Yes, please feel free to keep this introduction. I didn't realize I didn't write one. So thank you for taking care of that. 

-

dayleparker:
Hi Chris,

Thanks so much for your feedback and insights -- much appreciated!
I did a bit more digging and found another definition for providers from our engineering team, where the 'managers' are also classed as 'providers': https://access.redhat.com/documentation/en-us/reference_architectures/2017/html/deploying_cloudforms_at_scale/architecture#providers
 For this reason, I decided to leave the book subtitle as just 'providers' and the Abstract sentence to include 'system managers' (good term!), and then in the Preface, I replaced 'external management system' with 'external management environment', since this particular doc makes special note to define that phrase in a different way to how I used it. Finally, I'll keep my ears open for future discussions on providers/managers terminology as well. It's definitely a bit confusing.

I also went through and removed "Manager" from any phrase that was "Red Hat Virtualization Manager Provider", because agreed, that is totally confusing. (Same for SCVMM, but it's a bit harder as it's the product name). Thanks for noticing that and pointing it out!

Anyway, let me know if you think those changes are all right (or if more fixes are needed), and if you're happy with them, I think it should be ready to be merged.

New preview: http://file.bne.redhat.com/~dayparke/CloudForms/Providers3/build/tmp/en-US/html-single/
Cheers,
Dayle

cbudz:
Changes look good. thanks Dayle. Merging this request. 

-Chris

PR #372, opened 2017-05-22
Changed guide name to High Availability Guide

dayleparker:
This guide has changed scope and now includes administration instructions (and more to come).
Following docs team discussion, we'll rename it the High Availability Guide. This PR covers the renaming.


dayleparker:
Double checked all all looks OK. Merging.

PR #360, opened 2017-05-15
[api] API v2.4.0 Documentation

abellotti:
[api] API v2.4.0 Documentation

- Updated Versioning page to talk about v2.4.0 and its features
- Added the newly exposed collections in api/reference/collections.adoc
- Added /api/actions reference
- Added /api/conditions reference
- Added reference for /api/alerts, /api/alerts/:id/alert_actions and /api/alert_definitions
- Added /api/authentications and /api/configuration_script_payloads/:id/authentications reference
- Added /api/configuration_script_sources reference
- Querying load_balancers as subcollection or providers or instances
- Ability to copy Orchestration Templates
- Added CRUD support for /api/policies
- import_vm action on /api/providers/:id
- Ability to copy Service Orders
- Querying orchestration_stacks of services
- Added CRUD on /api/service_dialogs including service dialog copy
- Adding support for bulk assign and unassign of tags for vms and services
- Adding and Removing resources from services
- Removing all resources from services
- Adding and Removing approvers for service_requests
- CRUD on /api/service_templates
- Editing Basic VM Info.
- Added Snapshot support for Vms and Instances
- Updated resource_attributes.adoc to reflect the new collections
- Updated providers attributes

PR #376, opened 2017-05-25
[api] Specifying Connection Configurations on a Provider

jntullo:
Recently there have been multiple BZs that have been solved by demonstrating the use of connection configurations, as well as questions in the API room about them. Some of these questions are likely avoidable with the correct documentation.

This update provides two examples of using `connection_configurations` that may be useful.

Thoughts @abellotti ?


abellotti:
This is great, the more examples, the better.  

would like https://github.com/ManageIQ/manageiq_docs/pull/360 merged first, then this.  Thanks.

abellotti:
With #360 merged (thank you @adahms), this PR is next.  @adahms can we borrow your 👀  when you get a chance ?

adahms:
@abellotti - Looks good to me; merging!

PR #379, opened 2017-05-30
Docs for self-signed CA workarounds

dayleparker:
Hi Suyog,

I've added an appendix on adding a self-signed CA, and notes in OpenStack and SCVMM sections linking to this appendix, as they seem to be the only providers affected by this now (ie. there is SSL validation in the UI, but no option for a self-signed cert). 
Please let me know if you have questions or comments -- I'll apply any SME feedback later so we can get this in for GA.

https://bugzilla.redhat.com/show_bug.cgi?id=1448711

http://file.bne.redhat.com/~dayparke/CloudForms/SSLCAcert/build/tmp/en-US/html-single/#app-self_signed_CA

Many thanks!
Dayle

PR #380, opened 2017-05-30
Implemented peer review suggestions from PR #379

dayleparker:
Hi Suyog,
Thanks for your work reviewing and merging the content about self-signed SSL certs in https://github.com/ManageIQ/manageiq_docs/pull/379. I think your suggestion to combine the points into Adding an OpenStack Infrastructure Provider neatens up the content, so I've combined those two notes. I also edited the note in the appendix section so it's a little more clear.

New preview: http://file.bne.redhat.com/~dayparke/CloudForms/SSLCAcert2/build/tmp/en-US/html-single/#app-self_signed_CA

Please let me know what you think!
Thanks again,  
Dayle


suyogsainkar:
@dayleparker Hi Dayle, Awesome - thanks for implementing the suggestions; looking good. Merging to master.

PR #383, opened 2017-06-01
Added extra step to update database schema

dayleparker:
Hi Suyog,

An additional comment came in after the new content was merged for https://bugzilla.redhat.com/show_bug.cgi?id=1442350 -- Felix suggested adding a step (now step 7) that would solve issues faced in some previous errata. I've added this in now (having originally removed it) -- would you mind doing a quick sanity check on this additional step, and merging if it looks all right? Apologies for the extra review request.

http://file.bne.redhat.com/~dayparke/CloudForms/HAupdate2/build/tmp/en-US/html-single/#updating-ha

I'll cherry-pick to euwe after as well.
Thank you,
Dayle

### DIFF
--- a/doc-High_Availability_Guide/topics/Updating_HA.adoc
+++ b/doc-High_Availability_Guide/topics/Updating_HA.adoc
@@ -59,6 +59,14 @@ This step is not required on the database-only appliances, as `evmserverd` does 
 # yum update
 ------
 +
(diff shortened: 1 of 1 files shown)


suyogsainkar:
@dayleparker Thanks for adding this step! Looks good to me; merging.

PR #385, opened 2017-06-02
Added note about obtaining a self-signed CA from Openshift Hawkular

dayleparker:
Hi Suyog,

After some discussions with CloudForms engineering and some members of the OpenShift and CFME SBRs, I've added a note to the Hawkular endpoint section about obtaining the self-signed SSL certs to make it easier for users.

There might be future work on this, but I think this is all that's needed on this bug for now. If any more SME feedback comes in on the topic, I'll address it in a new BZ.

http://file.bne.redhat.com/~dayparke/CloudForms/SSL-customCA/build/tmp/en-US/html-single/#adding_openshift_provider

https://bugzilla.redhat.com/show_bug.cgi?id=1457036

Many thanks Suyog! Please let me know if you have questions. (Note to self -- this is only for 4.5)
Cheers,
Dayle


suyogsainkar:
@dayleparker Hi Dayle, I think the newly added note (along with the example) reads well and should make it clear for the user. Merging to master. Thanks -Suyog

PR #395, opened 2017-06-06
Added a note about supported SSL options for Hawkular.

adahms:


JStickler:
@adahms, LGTM as long as we don't mention Hawkular in the downstream docs.

PR #429, opened 2017-07-12
[FINE] Bz1469652 fine

cbudz:


cbudz:
Backporting to Fine based on PR425

PR #430, opened 2017-07-12
[FINE] update VNC section with additional info

cbudz:


cbudz:
Backporting to FINE from PR410

PR #440, opened 2017-07-17
include force option information

cbudz:
Hey Dayle,

Please see here my additions regarding the force option when creating a backup. 

-Chris


dayleparker:
@cbudz Hey Chris, 
The addition looks great. There are two really small things I wasn't sure about (apologies for the  nitpick!):
* Should it be "an instance" in this sentence instead of "the instance" ["Select Force? to allow backup of a volume attached to the instance. "]? I re-read the section a few times and it didn't seem like it was talking about a particular instance; it just read a little strange to me.
* This is very small -- but I think we generally mark up a status from the GUI with some kind of Asciidoc markup -- so instead of "available" and "in-use", it would be more conistent with the guide to mark up as *available* (GUI label markup) or `available` maybe. 
Thanks and please let me know what you think.
Cheers,  
Dayle

cbudz:
Hey Dayle,

Thanks for catching those items. I've marked up those status selections per GUI label markup from the **Cloud Volumes** summary page.
-Chris

dayleparker:
Looks good to me, Chris! Thanks for making those updates :) Merging now...
Cheers, Dayle

PR #441, opened 2017-07-18
Bz1467502 amazon catalog master

suyogsainkar:
https://bugzilla.redhat.com/show_bug.cgi?id=1467502
https://bugzilla.redhat.com/show_bug.cgi?id=1467503

### DIFF
--- a/common/create-amazon-catalog-item.adoc
+++ b/common/create-amazon-catalog-item.adoc
@@ -1,5 +1,5 @@
 
-Create a catalog item that uses an Amazon
+Use the following procedure to create an Amazon catalog item. Once created, the catalog item joins the service dialog with all of the options in the provisioning dialog; then, users can order Red Hat Enterprise Linux instances from the self-service catalog in the Service UI.
 
 . Navigate to menu:Services[Catalogs], then click on the *Catalog Items* accordion. 
 . Click image:1847.png[](*Configuration*), then image:1862.png[](*Add a New Catalog Item*).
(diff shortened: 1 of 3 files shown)


common/create-amazon-catalog-item.adoc
```diff
@@ -0,0 +1,16 @@
+
+Use the following procedure to create an Amazon catalog item. Once created, the catalog item joins the service dialog with all of the options in the provisioning dialog; then, users can order Red Hat Enterprise Linux instances from the self-service catalog in the Service UI.
```
cbudz: I would recommend taking the second sentence here and breaking it into two sentences since it exceeds the recommended 25-word limit for a sentenced in the IBM guide. My only other recommendation here would be to replace the verb _join_ with a synonym like _combine_ or _link_. I had to read it a few times before I realized what it was conveying, but that could be me...I tend only to recognize the verb _join_ in a software context when referring to tables in a database.

suyogsainkar:
Hey @cbudz, thanks for your comment. The introduction text certainly needed some tidying up, so I avoided changing the bug status to ON_QA but accordingly missed leaving a comment for you in the PR :) But thanks for your prompt review. I've reworded the introduction line to implement your suggestion. Let me know what you think. By the way, I might make a few minor updates based on some further testing, but will open a separate PR in that case.

cbudz:
Thanks, Suyog. All looks good. Merging this PR.

PR #458, opened 2017-07-26
Updated sections on updating appliances

dayleparker:
I've revised the sections on updating CloudForms and updating the appliance (with collaboration from Andrew Spurrier, John Hardy, and Tae Ho Choi) after feedback from several teams.

The main issues highlighted by the conversations are:
* The term "appliance operating system", which we previously used, is confusing. Changed to "appliance".
* Appliance updates can be done via yum or Satellite, and are supported by Red Hat. However, if you have a heavily customized CloudForms environment, things may break. Back everything up before applying updates. 
* Downtime is required for updates.
* Updating the CloudForms application (from inside the UI, which just updates CloudForms packages), and the appliance (done from the command line, useful for automation) are separate operations done in different ways. 

I had to do a bit of restructuring with the files (was too hard to make an xref link from a common section...) as well, and added a missing full stop in the 'Registering" section.

### DIFF
--- a/common/configuration-register-appliance.adoc
+++ b/common/configuration-register-appliance.adoc
@@ -10,7 +10,7 @@ To register your appliance with Red Hat Subscription Management or Red Hat Satel
 To configure registration for a region:
 
 . Log in to the appliance as the `admin` user.
-. From the settings menu, select *Configuration*
+. From the settings menu, select *Configuration*.
 . Select *Region* in the accordion menu and click the *Red Hat Updates* tab.
 . Click *Edit Registration*.
 . Configure registration details for the {product-title} appliance using one of two available options:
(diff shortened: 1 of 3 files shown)


dayleparker:
@suyogsainkar 
Hi Suyog,
I think this is ready for a final review by docs, as I think we've reached an agreement on the rewording. I've forwarded you the related email thread so you can see all the conversation that happened previously.

Please let me know what you think, and if it looks ready to merge. I know we have a Warning next to an Important box in one place, but I wasn't sure the best way to deal with that. Any ideas welcome, but happy to leave it like that as well to emphasize different parts. Hoping this section will become a bit better structured/less nested with a future General Config guide overhaul.

Thanks so much,
Dayle

suyogsainkar:
@dayleparker Hi Dayle, thanks for all the updates! I've finished reviewing the changes and everything looks good to me. As discussed, I'm merging this PR now, and you can make any further changes if required, based on Taeho's feedback.

PR #462, opened 2017-07-27
Bz#1467504: Creating a Generic Catalog Item 

cbudz:
Hey Suyog,

I've added the required procedure on adding a generic catalog item Provisioning VMs guide, which you can find in the included files. While working on this, I did make an additional change of moving this item, and those for Ansible Playbook and Ansible Tower catalog items up a level in the index. My rational is that, since each of these is really its own unique workflow, they shouldn't be nested as subtopics under Creating a Catalog Item (that deal with different providers).

Let me know what you think.

-Chris

### DIFF
--- a/common/create-generic-catalog-item.adoc
+++ b/common/create-generic-catalog-item.adoc
@@ -1,0 +1,21 @@
+. Navigate to menu:Services[Catalogs].
+. Click the *Catalog Items* accordion.
+. Click image:1847.png[](*Configuration*), and then image:1862.png[](*Add a New Catalog Item*).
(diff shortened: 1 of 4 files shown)


suyogsainkar:
@cbudz Hey Chris, the changes look good to me. I suspect there is going to be a bit of an overhaul to the structure of Catalogs and Services for which I've started some work on already as part of another bug, and there is more to come. This is for the same reason you've identified yourself that each has its own unique flow and needs to better match the customer's use case. Merging this to master. Thanks, Suyog

PR #468, opened 2017-08-04
add missing character to install instruction

cbudz:
Hey Dayle,

Quick one here. I've added the missing "." to the final command in the installation procedure.

-Chris


dayleparker:
Looks good to me, thanks Chris!

PR #473, opened 2017-08-08
[FINE] Removing 'deploy from external DB' section

dayleparker:
Hi Andrew,

I've removed 'Deploying the CloudForms Appliance Using an External Database' from the _fine_ branch as it's not supported in CloudForms 4.5.1 (as we learned via feedback from Satoe-san in QE). As this will be supported in CFME 5.8.2; left the content in master for publishing with next release.

Would you mind merging this PR please? I will republish the guide to the Portal afterward.

Also, I wonder if we should create a BZ to remind us to include the content again for 5.8.2, as having it in master will only get it into the upcoming CloudForms 4.6 guides, IIUC. What do you think?

Thank you,
Dayle


adahms:
@dayleparker - Looks good to me; merging.

Good call on having a tracker to reinstate this content when ready. I'll raise one and note the commit ID so that we can just run a 'git revert'

dayleparker:
@adahms thank you for merging, and creating the tracker. Feel free to assign it to me as well :)

PR #488, opened 2017-08-18
Bz1477927

cbudz:
Hi Dayle,

I've added some information about debugging verbosity in the note that begins this topic, as well as a step to the procedure for both provisioning and retirement. Let me know what you think.

-Chris


dayleparker:
The new content looks good to me, Chris. Thanks for providing these details!
Cheers,
Dayle

PR #526, opened 2017-10-02
Fix conflicting instruction on minimum requirements

FlorianOnmyown:
Fix conflicting instruction on minimum requirements


adahms:
@FlorianOnmyown - Thanks for the PR!

Looks good to me; merging.

PR #536, opened 2017-10-16
[BZ1455568] Add Enable Server Role procedure and add note about establishing network id

cbudz:
Hi Dayle,

Here's the PR on enabling the Ansible Server role and a note admonition calling out best practice of first setting up the network ID for the appliance. Let me know what you think.

-Chris

### DIFF
--- a/doc-Managing_Providers/topics/Automation_Management_Providers.adoc
+++ b/doc-Managing_Providers/topics/Automation_Management_Providers.adoc
@@ -49,6 +49,12 @@ A table of all workers and current status will appear from which you can confirm
 === Adding a Playbook Repository
 Add a repository so that {product-title} can discover and make available your playbooks.
  
+[NOTE]
+
+====
+Configure your {product-title_short] appliance network identity (hostname/IP address) before adding Ansible playbook repositories. Restart the `evmserverd` service after changing the hostname or IP address to ensure repositories add properly.
+====
+ 
 . Navigate to menu:Automation[Ansible > Repositories].
 . Click *Add*.
 . Provide a Repository Name in the *Name* field. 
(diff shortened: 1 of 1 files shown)


dayleparker:
Hi Chris,
The content and structure of the update look good

A few comments before merging:

1.  It would be good to make this sentence a little clearer -- "In Red Hat CloudForms, the Ansible role is disabled by default. Enable the server role to utilize Ansible Automation Inside. "

- ie. Is there a server role called "Ansible" as well as "Embedded Ansible"?

Suggested edit, if this is correct: 
"In Red Hat CloudForms, the **Embedded Ansible** role is disabled by default. Enable this server role to utilize Ansible Automation Inside. "

2.  It looks like this may be from previous releases, but I would suggest a numbered list instead of 4 bullet points in the 'workflow' after the sentence "Ansible is built into CloudForms so there is nothing to install. The basic workflow when using Ansible in Red Hat CloudForms is as follows: ".  What do you think? I gather the order of steps is pretty important here.

3.  I do a lot of work in the General Configuration guide, and couldn't find any mention of the Embedded Ansible server role in the reference table or guide itself. I would suggest adding a description there, as whenever I see "server role" and want to know more, I go to that guide. Up to you whether it would be a fix for a new bug (that might be better?) or as part of this branch... or is it documented elsewhere already? I just looked in the 4.5 version in 4.1.4.1.3. Server Roles:
https://access.redhat.com/documentation/en-us/red_hat_cloudforms/4.5/html-single/general_configuration/#regions

Also in the General Configuration guide, there are lists of "region-aware" and "zone-aware" server roles, so not sure if Embedded Ansible applies to either of those categories, and should be listed in those tables. ?

Cheers,  
Dayle

carbonin:
> Is there a server role called "Ansible" as well as "Embedded Ansible"?

Nope, the role name is "Embedded Ansible"

> Also in the General Configuration guide, there are lists of "region-aware" and "zone-aware" server roles, so not sure if Embedded Ansible applies to either of those categories, and should be listed in those tables. ?

If I understand the table correctly I think Embedded Ansible would be region aware and Primary/Secondary

cbudz:
Hey @dayleparker,

I've made the recommended revisions per items # 1 and 2 in your above comment. I broke out item 3 into its own BZ, on which I added you as the QA contact. 

-Chris

dayleparker:
Hey Chris,
Thanks for making those edits and the new BZ -- it looks great to me.
Merging!
Cheers,
Dayle

PR #537, opened 2017-10-18
…

dayleparker:
Hi Suyog,
I've added an image showing roles-user-groups relationship, and edited those sections to provide more details about limitations and relationships, and how to find out more info about roles in Product Features.

I also confirmed with Joe Vlcek that assigning multiple groups to a user is a new feature for 4.6, so the cherry-picking of this change will have to be manually done (note to self).

Can you please let me know what you think of the changes? https://bugzilla.redhat.com/show_bug.cgi?id=1413495#c4 for reference.

Many thanks,
Dayle


suyogsainkar:
@dayleparker All updates look good to me; merging to master. Thanks, Suyog

PR #540, opened 2017-10-18
[BZ1499106] Create Template from VM

cbudz:
Hey Dayle,

I've set-up this PR related to BZ1499106. It includes the workflow to create a template from a RHV based VM. Please let me know what you think and what revisions you might suggest.

Thanks!
Chris


dayleparker:
Hey Chris,  
The procedure looks really great - easy to follow and with well-placed conceptual descriptions right where I was wondering 'what does that term mean?' (Defining 'sealing' was really helpful too). I tested this procedure easily on our appliance :) I just noticed a few small punctuation-type things that might need editing, and there was just one line I found a little confusing:

* The VM must be associated with a Red Hat Virtualization Provider.  -- "provider" should be lowercase

* This line is a bit confusing to me -- "The Red Hat Virtualization provider must support publish vm.":
- Is "publish vm" a system call? Or a parameter maybe?
- I think we should conditionalize "ovirt-engine-4.0" for MIQ, and Red Hat Virtualization 4.0 for CFME downstream. (I confirmed with Tahlia from the RHV team that ovirt-engine-4.0 is the same as RHV 4.0)
- I think we would probably mark up a package name and filename with italics (might be worth checking the style glossary) for ovirt-engine-4.0 and settings.yml. 
- Where is the file "settings.yml" located -- ie. a file on the RHV provider?

* Generally we always used to spell out "VM" as "virtual machine". I think it was so it wouldn't be confused with "virtual memory". Or maybe it was to be more clear for translation? I'm not sure if that's a hard rule anymore, but it might be good to write one of the first mentions of "virtual machine (VM)" with the abbreviation like that. However, I also imagine "VM" is a nice searchable term for customers ... what do you think? 

* Should "Create a template based on an existing virtual machine:" start with "To create..." ?

* I recommend adding comma in step 5 after "Under the Request tab"

* Step 9 is missing punctuation at the end of the phrase.

Please let me know if you have any questions!
Cheers,
Dayle

cbudz:
Hey Dayle,

I've made some revisions to the file, which you can see in the preview. I've made the following changes based on your feedback:

**This line is a bit confusing to me -- "The Red Hat Virtualization provider must support publish vm." Is "publish vm" a system call? Or a parameter maybe?**

- I removed reference to this entirely, since users need a 4.0 provider regardless and a setting file configured to 'true'. I've added this below. IN RHV its referred to as "make template" but CFME calls the procedure 'publish', whch is how it was referenced to me. 

**I think we should conditionalize "ovirt-engine-4.0" for MIQ, and Red Hat Virtualization 4.0 for CFME downstream. (I confirmed with Tahlia from the RHV team that ovirt-engine-4.0 is the same as RHV 4.0)
I think we would probably mark up a package name and filename with italics (might be worth checking the style glossary) for ovirt-engine-4.0 and settings.yml.**

- I added conditional parameters to both the miq and cfme files in the attributes file for product names. They are currently set to Red Hat Virtualization for CFME and oVirt Engine for MIQ. I've formatted to follow style guidelines for a path and removed reference to the package entirely, in favor of referring to the products. 

**Where is the file "settings.yml" located -- ie. a file on the RHV provider?**

- Since the setting file is on the appliance, I added the full path, which should direct the user to the file on their appliance.

**Generally we always used to spell out "VM" as "virtual machine".**

- I checked our Red Hat conventions, and our guideline is to spell out the full term then follow using the abbreviated VM, so I've just made some changes to follow that policy. 

-Chris

PR #542, opened 2017-10-19
added new screenshot for multiple groups feature

dayleparker:
Hi Suyog,

This is a follow-up to https://github.com/ManageIQ/manageiq_docs/pull/537 -- I just realized while cherry-picking (manually to 4.5) the screenshot needed to be replaced also for the multiple groups to a user feature, and I totally missed that before.

Would you mind having a quick look at the screenshot and letting me know if anything's out of place, or it looks ok?

Thanks again!
Cheers,
Dayle


suyogsainkar:
@dayleparker Ah, ok - thanks for adding that, Dayle. It looks good to me; merging. Cheers, Suyog

dayleparker:
Thanks Suyog!

PR #543, opened 2017-10-23
Conditionalized the Building Images on OpenShift section

dayleparker:
Hi Chris,
This is a really quick one -- I've conditionalized the "Building Images on OpenShift" section so it only builds for the MIQ version of the guide.
Thanks for reviewing and merging!
Cheers,
Dayle


cbudz:
Looks good, thanks Dayle. Merging this request.

PR #514, opened 2017-09-24
Bz1488260 azure updates

suyogsainkar:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1488260

PR #545, opened 2017-10-24
[FINE] Modified the steps in downstream only for the CF 4.5.2 release.

suyogsainkar:
Modified the steps wherein the conversion and alignment steps aren't required with this release build anymore. (Currently in downtream 5.8.2 only as advised by engineering). 
Also updated the Install the Azure Command-Line Interface steps as suggested which does not use curl. All of these updates are tech reviewed and approved by the SME. 
Note that some of the other improvements currently under progress as part of the master branch (https://github.com/ManageIQ/manageiq_docs/pull/514) will be backported to 4.5 as soon as they're reviewed and approved.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1488260
Preview: http://file.bne.redhat.com/ssainkar/4.5.2/cf-on-azure/html-single/

Thanks!
Suyog

### DIFF
--- a/doc-Installing_on_Microsoft_Azure/topics/installation.adoc
+++ b/doc-Installing_on_Microsoft_Azure/topics/installation.adoc
@@ -29,6 +29,7 @@ ifdef::cfme[]
 . From the list of installers and images, select the {product-title} appliance specified for Microsoft Azure download link.
 endif::cfme[]
 
+ifdef::miq[]
 [[uploading-the-appliance-to-microsoft-azure]]
 === Uploading the {product-title} Virtual Appliance to Microsoft Azure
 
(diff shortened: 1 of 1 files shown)


suyogsainkar:
@adahms Hi Andrew, please would you mind reviewing and merging the updates if the changes look and render OK for the downstream version of the doc for the 4.5.2 release? Let me know if you spot anything that needs fixing. Thanks! - Suyog

suyogsainkar:
@cbudz @dayleparker Hey Chris, Dayle, would any of you mind reviewing and merging these updates in relation to the 5.8.2 release? Please see my notes above. Let me know if you spot anything that needs fixing. Many thanks, Suyog

dayleparker:
Hey Suyog,

The new content looks really good. In the interest of time, I've merged it (@cbudz, hope that is OK - feel free to comment though!) I have also left the bug open since you have lots of other work in progress (but will note that this PR has been merged). 

I have a few suggestions and very small things that came to mind that you might want to consider when you're doing the other changes for this bug in the master branch. Feel free to address them (or not) as you see fit. Nice work!

Cheers,
Dayle

* I believe it would be best to change the full stop at the end of steps and sentences pointing to lists to a colon, to keep consistent with our other docs. I think it's more correct... For example: 

" You can upload the appliance to an Azure environment using the following two methods:"
or 
"Import the Microsoft repository key:"

* What do you think about adding xref links to the sections describing the two options under "You can upload the appliance to an Azure environment using the following two methods. "? I understand this might complicate things by having to add a mention here that you must install the Azure Command-Line Interface first though -- that's a prerequisite for both ways of uploading the appliance, right? Just a thought.

In "_1.2.2. Uploading the CloudForms Virtual Appliance Using Azure Powershell Script_":

* Should this be "Using the Azure Powershell Script" (or "an Azure Powershell script"?) Same for "Using Azure Command-Line Interface (Azure CLI) ", and the section titles for 1.2.2 and 1.2.3.

* For the first note in this section, I went to the Azure link and searched for the "To install the cmdlets" section but couldn't find it. Any idea if we should be pointing elsewhere? Also I think stylewise, a section name should be in italics. (I could be wrong)

* In step 2, I wonder if "Additionally, create the storage container defined in "BlobDestinationContainer" in advance. " should instead be its own step before this step. Could we provide instructions for this, name it as a prerequisite, or point to the right instructions in the Azure docs?

* In step 3 - "Then, define your VM and VHD name, your system/deployment name and size. Next, you will set the appropriate Storage, Network and Configuration options for your environment. " -- these are all done as part of the script where you create your virtual machine, right? If that's correct, I suggest editing the wording a little so it doesn't sound like an additional step, maybe something like "Create a script to configure your virtual machine. The script defines your VM and VHD name....storage, network, .... for your environment." What do you think?

* Do you need to also run the script to create the VM, or is that covered elsewhere?

In section "_Uploading and Provisioning the CloudForms Virtual Appliance Using Azure Command-Line Interface"_:

* In step 1, should the "Note" be a note box? Good idea putting that tip there :)

suyogsainkar:
Hi Dayle, thank you for merging and making those suggestions. I will take a look and incorporate as needed as part of my ongoing work. Cheers, Suyog

Issue #573, opened 2017-11-06
Broken link to  See Database Requirements


mfeifer:
http://www.manageiq.org/docs/reference/latest/doc-Installing_on_Red_Hat_Virtualization/miq/
Section 1.3, step 4d has broken link

PR #577, opened 2017-11-07
Fixed a broken link in the section on running CloudForms in RHV.

adahms:


adahms:
@mfeifer - This PR addresses #573 

With this change, I have conditionalized the link so that a correct upstream and downstream version appear.

Would you be able to take a look and let me know what you think? If the changes are good, I will backport them to the 'fine' branch.

doc-Installing_on_Red_Hat_Virtualization/topics/installation.adoc
```diff
@@ -196,7 +196,8 @@ The virtual machine is created. To view the virtual machine, select the data cen
 .. To add a disk, select your virtual machine from the *Virtual Machines* tab.
 .. From the details pane, click the *Disks* tab. 
 .. Click *New* to open the *New Virtual Disk* dialog.
-.. Specify a *Size* in GB for the disk that allows sufficient space for your database. See https://access.redhat.com/documentation/en-us/red_hat_cloudforms/4.6-beta/html/deployment_planning_guide/introduction#database-requirements[Database Requirements] in the _Deployment Planning Guide_ for size considerations.
+ifdef::cfme[.. Specify a *Size* in GB for the disk that allows sufficient space for your database. See https://access.redhat.com/documentation/en-us/red_hat_cloudforms/4.6-beta/html/deployment_planning_guide/introduction#database-requirements[Database Requirements] in the _Deployment Planning Guide_ for size considerations.]
```
chessbyte: isn't this upstream documentation?

adahms: @chessbyte - That is correct, and because this content is shared between both the upstream and downstream, we can use a conditional like that above so that this version only appears in the downstream, and the version below only appears in the upstream. This PR helps make this distinction.

mfeifer:
👍

PR #590, opened 2017-11-21
add info on TTL to playbook catalog item creation procedure

cbudz:
Hey Suyog,

I've added some info here regarding the new TTL field for ansible playbook catalog items. I decided to keep the description brief, and in line with the procedural step rather than calling it out in a note. Also, I kept the description rather brief, as I think this is more or less all the user needs to know. 

Let me know what you think.

-Chris

### DIFF
--- a/common/create-playbook-service-catalog-item.adoc
+++ b/common/create-playbook-service-catalog-item.adoc
@@ -19,7 +19,7 @@
 .. Assign the appropriate *Machine Credentials* from the drop-down list. 
 .. Add *Cloud* or *Network Credentials* from the drop-down lists. 
 .. Choose the *Host* against which to run the service item. 
-.. Set the *Max TTL* in minutes. The *Time To Live (TTL)* field allows you to set the maximum execution time for the playbook to run. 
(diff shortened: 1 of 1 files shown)


common/create-playbook-service-catalog-item.adoc
```diff
@@ -19,6 +19,7 @@
 .. Assign the appropriate *Machine Credentials* from the drop-down list. 
 .. Add *Cloud* or *Network Credentials* from the drop-down lists. 
 .. Choose the *Host* against which to run the service item. 
+.. Set the *Max TTL* in minutes. The *Time To Live (TTL)* field allows you to set the maximum execution time for the playbook to run.
```
suyogsainkar: @cbudz Hey Chris, thanks for the addition. I think the brief description in the procedure reads well and provides enough info to the user.  By the way, I was able to see the **Max TTL (mins)** field, but not **Time To Live (TTL)**. If Time To Live (TTL) is not a field on the UI, I'd omit the mark up. Or, should that also read **Max TTL**?

cbudz:
Hey @suyogsainkar thanks for the feedback. I removed the markup from Time To Live (TTL), and have updated the preview. 

-Chris

suyogsainkar:
@cbudz Looking good; merging to master.

PR #606, opened 2017-12-06
Bz#1519310 Removed master SAML URL

dayleparker:
Hi Chris,

I've removed the line instructing users to configure the "Master SAML Processing URL" from the procedure under the heading "BZ#1519310-masterSAMLURL", which looks like all that is needed for this particular bug - https://bugzilla.redhat.com/show_bug.cgi?id=1519310
(The linked support case might require more docs changes in new BZs however, it's still ongoing).

Would you mind reviewing please? Let me know if you think anything else is needed to be clear.
I've also made this change in the WIP Authentication Guide branch in Gitlab so that it doesn't go missing (thanks for all your review work on that guide btw!)

Cheers,
Dayle


cbudz:
Looks good, thanks Dayle. Merging this request.

PR #612, opened 2017-12-13
Bz#1523471 VM retirement features

dayleparker:
Hi Chris,

I've edited the section "6.1.4. Setting a Retirement Date and Time for a Virtual Machine or Instance" to reflect the new functionality in CloudForms 4.6 to set times and relative times. Please let me know if you think it's clear enough, or if you have any comments:

http://file.bne.redhat.com/~dayparke/CloudForms/VMretire/build/tmp/en-US/html-single/#setting-a-retirement-date-for-a-virtual-machine-or-instance

https://bugzilla.redhat.com/show_bug.cgi?id=1523471

Thanks!
Dayle


dayleparker:
Hi again Chris,
I noticed something while testing that I wasn't sure if it was a glitch in my environment, or a change in the workflow -- there seems to no longer be an option to retire virtual machines from Services > Workloads > Lifecycle. (I tried it on a few VMs on all of our 4.6 test instances). The path I was able to use to retire VMs was Compute > Infrastructure > Virtual Machines > Lifecycle. 

I've now changed the lines in the retirement chapter leading from Services>Workloads to the above. Hope that makes sense (and let me know if you find a way to make it work from Services). I think this is a change we need to make in 4.5 as well after testing there.  So that's the last commit now. :)

Thanks again,
Dayle

cbudz:
Hey Dayle,

This looks good and I didn't run into any problems understading how the feature was described. 

And yes, it appears the workflow for setting retirement is as you've marked in the docs. I didn't see the retirement option under Service>Workloads.

Merging this request. 

-Chris

PR #615, opened 2017-12-15
Bz#1523466 servicedialoghelp

dayleparker:
Hi Chris,

I've added an optional step about what to add to the "Help" field in the Service Dialog editor. Please let me know what you think -- if the second sentence seems unnecessary, I'm happy to delete it (or edit).

I also made a few edits where I saw the new Automation>Automate path hadn't been updated.

Thanks for reviewing,
Dayle


cbudz:
Hi Dayle,

I think this reads well. The second sentence is helpful in explaining to users exactly why an admin might want to take advantage of the optional field. Merging this request.

-Chris

dayleparker:
Thanks Chris! I appreciate the feedback :)

PR #620, opened 2017-12-18
s/postgresql94/postgresql95/

carbonin:
We've been on PG 9.5 for about a year.

There are also very similar instances of this issue in `doc-Appliance_Hardening_Guide/miq/build/en-US/master.xml`. Should I change those also or are those files generated by some build process?


adahms:
@carbonin - Thank you for the pull request!

Looks good to me; merging. The reference to master.xml is indeed for a file that does not need to be included, and I have now removed this.

PR #628, opened 2017-12-20
[BZ1523005] add procedure on ansible playbook method

cbudz:
Hi Dayle,

Would you mind taking a look at this PR? I've included a link to the preview in the BZ.

Thanks,
-Chris


dayleparker:
Hi Chris,
The procedure looks good. I ran through it and it was pretty easy to follow, and I really like that you explained certain terms in the relevant steps (like TTL). My knowledge of Automation/Playbooks is not super great, so I'm not sure what level of knowledge the person running the task would have before running this (you can choose to disregard comments below that are too n00b to be relevant!). I noticed a few small things that need a fix while running through the procedure on our .181 instance:

- In the preview, section 1.15.3's first {product_title}  didn't render in the preview, but the code looks all right in the source. It's probably not an issue but just wanted to mention it.

- Note: Our convention for guide titles in CloudForms "in Managing Providers" (and the guide name is in italics) instead of "in the Managing Providers guide", to be a little more brief I think was the reasoning.

- Should "Create a playbook automate method:" be "To create a playbook automate method: " ?

- Step 1 - I think the UI navigation needs editing to Automation > Automate > Explorer. 
 Also, I didn't know at first where to find the class to find the Methods tab, but found it after a bit of clicking around. Do you think it would be worthwhile to add something like navigate to a domain and then click something in that domain? I can see we sometimes use this format in the other procedures in this guide, and sometimes not. What do you think?

- Step 3 - Suggest adding "then" before the comma, to read more smoothly.

- Step 6.d. Cloud Credential is singular in the UI (.181)

- Step 7 - localhost is capitalized in the UI. (.181)

- Step 10 - I had a little trouble comprehending this one - would it have its intended meaning if it was reworded a little, maybe like "Use the Escalate Privilege toggle switch to enable user privilege escalation if credentials are called for during the playbook run. " ?

- Step 12 - The icon needs updating -- it's a little blue plus sign now, and I suggest removing "the" from before the icon. Also typo  - "paramtrs". Are the input parameters something the user must put in when running the playbook? Should it be explained here, or is it common knowledge for someone performing this task?

Thanks for reviewing these changes!
Cheers,
Dayle

cbudz:
Hi Dayle,

Thanks for the feedback. I've updated the doc with your suggestions. As related to your final comment, I've added a note to better explain what input substitution involves, and direct users to the Ansible doc. I've also got a question out to the developer, and when I hear back with any further required information, I can patch it into this section, since I need to update the Methods section with some content on this new feature as well as some other info.

-Chris

dayleparker:
Hey Chris,

It all looks great to me -- thanks for making those edits. Should I go ahead and merge this now (and verify the bug), or should I wait until you hear back from the developer about your question?

Cheers,
Dayle

dayleparker:
Awesome, thanks Chris. All merged!

PR #631, opened 2018-01-04
edited catalog item type step with new functionality

dayleparker:
Hi Chris,

I think you've been working with catalogs lately... would you mind reviewing the changes I made and letting me know if the wording makes sense for the new functionality (described in https://github.com/ManageIQ/manageiq-ui-classic/pull/2908), for docs BZ https://bugzilla.redhat.com/show_bug.cgi?id=1526554.

I also removed an old screencap from the step which I don't think was really needed, and a few other minor edits to the section. The previous version can be seen in step 4 in section 5.4.2 - https://access.redhat.com/documentation/en-us/red_hat_cloudforms/4.6-beta/html-single/provisioning_virtual_machines_and_hosts/#creating-a-catalog-item

Thanks for reviewing! Please let me know what you think.
Cheers,
Dayle

### DIFF
--- a/common/catalog-item-creation.adoc
+++ b/common/catalog-item-creation.adoc
@@ -1,7 +1,7 @@
 . Navigate to menu:Services[Catalogs].
 . Click the *Catalog Items* accordion.
 . Click image:1847.png[](*Configuration*), and then image:1862.png[](*Add a New Catalog Item*).
-. Select the *Catalog Item Type* you are adding. Only providers available to the appliance show in this list. The dialogs that appear will be filtered based on the selected type of provider. 
+. Select the *Catalog Item Type* you are adding. This list only shows items related to providers available or options activated in the appliance (for example, *Ansible Playbook* is available as a *Catalog Item Type* option if the Embedded Ansible server role is enabled on the appliance).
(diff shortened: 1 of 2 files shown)


Loicavenel:
@dayleparker I am not sure this "Only providers available to the appliance show in this list. The dialogs that appear will be filtered based on the selected type of provider." makes sense because some options like Orchestration are depending of multiple providers. And Ansible Playbook is not provider dependant but enable when a appliance role is active. May be something like: "This list will show only items related to providers available or Options activated like Ansible"

dayleparker:
@Loicavenel Thanks for your feedback -- I wasn't sure about the wording myself but that makes more sense.
I've now edited it so it's: "This list only shows items related to providers available or options activated in the appliance (for example, Ansible Playbook is available as a Catalog Item Type option if the Embedded Ansible server role is enabled on the appliance)." I hope that's correct/easier to read now :)

Loicavenel:
@dayleparker looks good.

cbudz:
Hey Dayle,

Everything looks good. Thanks for making those updates. Merging this request.

-Chris

PR #632, opened 2018-01-09
Bz#1517838 webservices roles

dayleparker:
Hi Chris,

I've updated the server role descriptions in the General Config guide to help customers with the issue Robb has seen (as reported in the bug); I've also shortened the product names in that table for better readability.

section 4.1.4.1.3. Server Roles-->
http://file.bne.redhat.com/~dayparke/CloudForms/Serverroles/GenConfig/build/tmp/en-US/html-single/#servers

I also added a Roles section (linking to the Gen Config guide)  to provide an overview of roles and their possible interdependency. Please let me know what you think.

http://file.bne.redhat.com/~dayparke/CloudForms/Serverroles/Deployment/build/tmp/en-US/html-single/#roles

https://bugzilla.redhat.com/show_bug.cgi?id=1517838

Thanks so much,
Dayle


cbudz:
Hi Dayle,

These changes look good, and the additionaly 'roles' section in the planning guide is help. Thinking about interdependency of server roles, I'm wondering if it wouldn't be helpful in the future to add a note or important admonition column in the Server Roles table to more immediately capture users' attention to which roles require orchestrated activitation to function properly. 

Merging this request.

-Chris

PR #678, opened 2018-01-22
Added section on hiding environment variables, screenshot

dayleparker:
Hi Suyog,
Would you mind reviewing this new section when you have a chance? Please let me know if you have any feedback.

Thanks so much,
Dayle


suyogsainkar:
@dayleparker Hey Dayle, thanks for the PR - looking good to me. I made two small edits as part of my peer review and committed the changes. Hope it makes sense. Merging to master. Cheers, Suyog

dayleparker:
@suyogsainkar works for me - thanks for reviewing and making those edits! Btw, I think perhaps the "Add" was when you made the role one way, and "Save" was for the other option, but I think it'll be clear either way for users.

suyogsainkar:
@dayleparker No worries. OK, cool - should be fine I think. Thanks for looking into it.

PR #688, opened 2018-02-01
add overview on playbooks service linking roles docs

cbudz:
Hi Dayle, 

Would you mind having a look at the files I've included here that cover the following:

- Service Linking: this is the primary request for the BZ
- Installing Roles: an inclusion that comes with the territory of working with roles.
- Optimizing Playbooks: this is based on the suggestion Andrew made in the BZ, about creating a new section on writing playbooks for use with CF.

Preview is included in the BZ

Thanks,
Chris

### DIFF
--- a/doc-Managing_Providers/topics/Adding_Ansible_Galaxy_Roles.adoc
+++ b/doc-Managing_Providers/topics/Adding_Ansible_Galaxy_Roles.adoc
@@ -1,0 +1,18 @@
+[[installing-ansible-roles]]
+
+=== Installing Roles on an Embedded Ansible Appliance 
+
+Roles are ways of automatically loading certain variable files, tasks, and handlers based on a known file structure. Grouping content by roles also allows easy sharing of roles with other users. 
+Install roles on a {product-title} appliance with the Embedded Ansible server role activated to optimize playbooks. 
+
+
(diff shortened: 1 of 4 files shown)


dayleparker:
Hey Chris,
It looks great overall -- I just spotted a few small things that I think need reviewing before we merge.
- In 3.1.7.,  compliment should be complement
-  In 3.1.7.2.- this sentence should end in a colon to link it to the steps below "Linking VMs back to the service that created it requires implementing the following tasks in the playbook used for provisioning:" Same with the steps in the example.
- I suggest naming the Example with what it's showing if it's going to be a header (for easier searching, and consistency). Like Example: Linking a virtual machine to a service (or similar) 
- I tripped over the term "href slug" - is that a common Ansible term? i looked it up and it sounded like it could be classed as an attribute, thoughts?
- Do you think we should add a link to the Ansible Galaxy documentation?
- I suggest removing "please" from the note - it's not needed.
- As step 3 is an alternative step, I think it should be part of step 2.

Please let me know your thoughts.
Thanks so much,
Dayle

cbudz:
Hi @dayleparker,

Thanks for the feedback. I've made changes based on each of the items you've listed above. 

Thanks,
Chris

dayleparker:
Looks excellent, Chris. Thanks for following up on this !
Cheers,
Dayle

PR #689, opened 2018-02-01
Bz#1534961 emailquota

dayleparker:
Hey Chris,
I've added some info about the necessity of an email address for provisioning in two spots in the guides. Screenshot previews are attached to https://bugzilla.redhat.com/show_bug.cgi?id=1534961.
Could you please let me know what you think?
Thanks so much,
Dayle


cbudz:
Hey @dayleparker.

Looks good to me, and I think the redundancy is called for in this case if omitting an email address can result in a request not provisioning. 

Merging this PR.

-Chris

PR #694, opened 2018-02-05
Update Access Restriction for roles UI name and details

dayleparker:
Hi Suyog,

This is a part 2 to a merged BZ (https://bugzilla.redhat.com/show_bug.cgi?id=1532411), where I came across a new UI name, which required a new screenshot in the GC guide and a little more explanation in a new section in Managing Providers.

I've updated both guides now, and updated screenshots in the General Config guide (not sure if we really need the screenshot in step 5, but I've updated it as it needed a refresh (and used to list Middleware as a category). A capture of the updated section is attached to the BZ, as file.bne is not working for me today: https://bugzilla.redhat.com/show_bug.cgi?id=1536911

Thank you,
Dayle


suyogsainkar:
@dayleparker Hey Dayle, thanks for all the updates - looks great to me. Merging to master. Cheers, Suyog

PR #703, opened 2018-02-09
Bz#1535567 update events

dayleparker:
Hi Chris,

Would you mind reviewing please?

There wasn't a screenshot to update as mentioned in https://bugzilla.redhat.com/show_bug.cgi?id=1535567, but I've gone through the list of events available in the UI via Control > Explorer > Events and added any that we hadn't documented to the list, and wrote descriptions based on similar event names:

* a few event groups added to "Table A.1. Event Types", based on what I could see from Control > Explorer > Events > [click on an event], then checking the Event Group

* Added any events visible in the UI (from Control > Explorer > Events) and descriptions.

For the 'pod container' ones, I learned that a pod can contain one or more containers, so hopefully the wording makes sense to you (let me know if not): https://docs.openshift.com/container-platform/3.7/architecture/core_concepts/pods_and_services.html

'Shelve offload' was also new to me but I used this for my definition: https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux_OpenStack_Platform/5/html/End_User_Guide/cli_manage_instances.html 

* I also reduced the long product names throughout the section to make the table briefer.

Let me know if you have any feedback!

Cheers,
Dayle

### DIFF
--- a/doc-Policies_and_Profiles_Guide/topics/chap_events.adoc
+++ b/doc-Policies_and_Profiles_Guide/topics/chap_events.adoc
@@ -45,6 +45,9 @@ checking of a condition.
 [cols=",",options="header",]
 |====
 |Event |Description
+
+|Container Image Analysis Request |NEEDINFO.
+
 |Container Image Analysis Complete |Check the condition when an analysis of a container image completes.
 
 |Container Image Discovered |Check the condition when a new container image is discovered.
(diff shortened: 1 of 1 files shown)


cbudz:
Hi @dayleparker,

Everything looks good, and I think the definition of 'selving' makes sense. Some nitpicky feedback I notced while reading through the events:

* *Provider Auth Incomplete Credentials* and *Provider Auth Unreachable* only state _for use with automate_ where the other events list '{product-name_short} automate'. For consistency do you think we should have all events descrubed the same way with the product name in there?

*  Orchestration Lifecycle - I'd replace the first intance of 'such as' with 'related to' to improve readability. 

Let me know what you think.

-Chris

dayleparker:
Hi Chris,
Thanks for reviewing, and these are great suggestions (and good catch on the 'related to', I must not have re-read that properly before sending it off for review!) 
I've made those changes now - please let me know if you spot anything else.
Cheers,
Dayle

cbudz:
Everything looks great, thanks @dayleparker.

Merging this request.

-Chris

PR #732, opened 2018-02-20
Custom Button Visibility and Enablement Filtering

cbudz:
Hi Dayle,

Here's the PR for the new procedure on adding expressions to a custom button that determine its visibility and enablement.

-Chris


dayleparker:
Hey Chris,
The new content looks good to me 👍 I made a couple minor fixes (missing navigation step, typo) but didn't want to hold up merging on account of these.

My only suggestion for a future fix is - I wonder how natural the terms in the section title are? (ie. is it searchable, would users use these terms?). Would it be better as something a bit more active along the lines of "enabling buttons and controlling button visibility" or "configuring button visibility and enablement" -- not that these are good ways to word it though, it's a tough phrase to shape. What do you think?

Cheers,
Dayle

PR #733, opened 2018-02-21
Updated the procedure to add OpenShift Containers Provider with infor…

suyogsainkar:
…mation to configure image inspector settings for OpenSCAP.

BZs: 
https://bugzilla.redhat.com/show_bug.cgi?id=1532450
https://bugzilla.redhat.com/show_bug.cgi?id=1532480

### DIFF
--- a/common/provider-ocp-add-container.adoc
+++ b/common/provider-ocp-add-container.adoc
@@ -58,6 +58,29 @@ In OpenShift, the default deployment of the router generates certificates during
 .. Enter the *Hostname* (or IPv4 or IPv6 address) or alert *Route*. 
 .. Enter the *API Port* if your Prometheus provider uses a non-standard port for access. The default port is `443`.
 .. Click *Validate* to confirm that {product-title_short} can connect to the alerts service.
+. Click the *Advanced* tab to add Image Inspector settings for scanning container images on your provider using OpenSCAP.
++
+[NOTE]
+====
+The Image Inspector extracts OpenShift container images from the registry and obtains the Common Vulnerabilities and Exposures (CVE) information (for effective scanning) via a proxy.
+====
++
+.. Enter proxy information for the provider in *HTTP*, *HTTPS*, and *NO Proxy*.
+.. Enter the *Image-Inspector Repository* information.
+.. Enter the *Image-Inspector Registry* information. 
+.. Enter the *Image-Inspector Tag* value. The default is `2.1`. Tag is a mark used to differentiate images in a repository, typically by the application version stored in the image.
+.. Enter the *CVE location*.
 . Click *Add*.
+
+[NOTE]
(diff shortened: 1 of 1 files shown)


suyogsainkar:
@moolitayer Hi Mooli, could you or someone on your team review the doc updates in the PR and provide feedback on the configuration steps and the contextual information in the note please? Thanks!

suyogsainkar:
@dayleparker Hi Dayle,

Would you mind peer reviewing and merging these updates if everything looks OK? I'll incorporate SME feedback, if any, post GA. Let me know if you spot any errors, or if you have any other feedback.

Thanks,
Suyog

dayleparker:
Hey Suyog,
The new content looks great and I compared it with the UI and it flows well.
Just a few small fixes before we merge:
* "Tag is a mark " --> should be A tag or The tag or The image-inspector tag....
* Image Inspector as a term should not be capitalized --> I believe it should be image inspector, the image inspector, or image-inspector. It's not documented very often (according to my searches), but I did find a useful tiger.io blog (maybe the one you were reading?) that used the term like :_"We have now configured everything to pull down the image-inspector container image from the local registry..." -- http://www.tigeriq.co/openscap-in-a-secure-environment/_ I found it interesting that the image inspector is actually a container image, and might be useful context for users to add somewhere in this section (it makes more sense to me now why you need to fill in all these registry details, etc). Image Inspector (in caps) sounds like a tool. 
* Just a small suggestion for the sentence in the note to avoid using the word via twice, how about rewording a little more actively, like:
"You can also set global default image-inspector settings for all OpenShift providers in the advanced settings menu by editing the values under `ems_kubernetes`, instead of setting this for each provider. For example:"

Please let me know your thoughts.
Cheers,
Dayle

dayleparker:
Thanks Suyog, it looks great! I'll move the two bugs to verified. Feel free to pass it on for review again in case you need to make changes after SME reviews.

suyogsainkar:
@dayleparker Hi Dayle, thanks for your review and feedback! Yes, that is the blog I was referring to which I also shared in our CF team meeting - lots of useful information in there. :+1:  I've applied those fixes in my latest commit and the PR is ready for reviewing/merging again. :) Cheers, Suyog

suyogsainkar:
Sounds good - thanks, Dayle. :)

moolitayer:
@suyogsainkar I'm now back from a 1 month PTO, FYI...

suyogsainkar:
@moolitayer Thanks for letting me know, Mooli. Hope you've had a good break.
I was wondering if you have any feedback to the updates in this PR, and to a related update in this currently WIP PR: https://github.com/ManageIQ/manageiq_docs/pull/768. As always, any and all feedback is welcome. Thank you!

PR #739, opened 2018-02-22
Added note on changing groups for quota settings

dayleparker:
Hey Chris,

I've added a note to managing quota section about changing groups (since this is where the 'gotcha' happened in the bug), but also added a screenshot and brief note in groups section, in case users are looking there for this info. (Screenshot because I thought the navigation is a little hard to describe, and not visible at all if the logged-in user doesn't have multiple groups assigned).

Please let me know what you think!
Cheers,
Dayle


cbudz:
Hi Dayle

Your additions look good and, in this instance (and after testing with the user account you created) I agree with including a screenshot for accessing the groups menu. 

Merging this PR.

-Chris

PR #762, opened 2018-03-07
fixed region/database step, added port step

dayleparker:
Hey Chris,
I've fixed the incorrect step that Krain has pointed out in https://bugzilla.redhat.com/show_bug.cgi?id=1520900, but also found when testing in 4.6 that now entering a port number is a step in this procedure, so I've added that too.

(Note, my testing didn't come across the step "You are prompted to create or fetch a security key.", but I'm reluctant to remove it as I was only testing this one procedure in isolation without multiple appliances, and Krain didn't see an issue with that -- so I'm hoping it's OK to leave that in)

Cheers,
Dayle

### DIFF
--- a/common/configuration.adoc
+++ b/common/configuration.adoc
@@ -111,8 +111,9 @@ endif::ssh[]
 All {product-title_short} appliances in a multi-region deployment must use the same key.
 ====
 +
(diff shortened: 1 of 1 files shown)


cbudz:
Hey @dayleparker 

Looks good, I'm merging this request.

-Chris

PR #763, opened 2018-03-09
Bz#1553485 vsphere

dayleparker:
Hi Suyog,

Would you mind reviewing the new content (it's based on the Google doc linked in the bug), and letting me know if it's clear, or needs any corrections?

Thanks so much,
Dayle


dayleparker:
Thanks so much, @suyogsainkar! I was vaguely familiar with the term IOPS from some previous work, so had overlooked expanding the acronym but should have. I think it's a great idea. I appreciate the fix! :)

PR #798, opened 2018-04-04
update provision openstack instance 

cbudz:
Hi Dayle,

I've made some updates to the section on provisioning an openstack instance to reflect the addition of volume and volume snapshots that can now be used. Please let me know what you think. Preview in the BZ. 

-Chris


dayleparker:
Hey Chris,
The revised content looks great. I made a small fix throughout (I'm hoping that's OK) to make the list of provisioning sources be images, volumes, and volume snapshots -- from my testing in the MBU appliance, I didn't see anything just called 'snapshot' or an 'image snapshot' -- that also lines up with the original bug, so hopefully it's right. (Feel free to unmerge and correct me if otherwise)
Thanks for your work on this!
Cheers,
Dayle

PR #804, opened 2018-04-10
Add details in add_disk method

cbudz:
Hi,

I've added some detail in sec 1.39 for the add_disk method.


suyogsainkar:
@cbudz Hey Chris, thanks for adding the info about the add_disk method along with an example in the table itself for now. Looks good to me; merging to master. Cheers, Suyog

PR #825, opened 2018-04-23
Add .yamllint from provider generator

jrafanie:
Add .yamllint from provider generator


cbudz:
Hi @jrafanie 

Where in the docs are you looking to add this file?

jrafanie:
@cbudz This repo only has a few yaml files so we can probably safely not use yaml lint.  I'll close this.

PR #828, opened 2018-04-26
[FINE] added admonition to link to smartstate/proxy Kbase solution

dayleparker:
This is only applicable to 4.5 - see https://bugzilla.redhat.com/show_bug.cgi?id=1541078


dayleparker:
A few more tweaks included with this in the second commit.

PR #830, opened 2018-05-01
Add note regarding host input value configurations

cbudz:
Hi Dayle,

I've added a note here in the ansible playbook custom button procedure related to the supported configurations for host value input. You'll notice the step that immediately preceeds the note addresses adding values for "specific hosts", then the note supplements that information. Please let me know if the details included in the note are clear and provide a distinction in how that info is provided (admin vs user). 

Thanks!
Chris


dayleparker:
Hi Chris,
Sorry for the delay in merging this!
The note reads really well and the placement makes sense to me, and it definitely helps clarify a tricky part in the UI. :) I think the distinction between user and admin-provided values makes perfect sense the way you wrote it. 
Cheers,
Dayle

PR #846, opened 2018-05-21
Added SELinux command in prepping section

dayleparker:
Hi Suyog,

Would you mind reviewing this addition and letting me know if you have any feedback?
I think it's more a preparation step than a prerequisites (the bug reporter suggested one of those two places). I added it to the end of the preparation procedure as all of those steps are done in OpenShift. Reading through the attached cases mentioned this needs enabling on each OpenShift node also.

Here's a preview of where it sits in this guide (https://access.redhat.com/documentation/en-us/red_hat_cloudforms/4.6/html/installing_red_hat_cloudforms_on_openshift_container_platform/installing-cloudforms):

setsebool

Cheers,
Dayle

### DIFF
--- a/doc-Installing_on_OpenShift_Container_Platform/topics/installation.adoc
+++ b/doc-Installing_on_OpenShift_Container_Platform/topics/installation.adoc
@@ -215,6 +215,13 @@ imagePolicyConfig:
 ----
 $ systemctl restart atomic-openshift-master
 ----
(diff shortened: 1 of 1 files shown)


suyogsainkar:
@dayleparker Hi Dayle, thanks for the PR - the new step and its placement looks good to me. Merging to master. Thanks, Suyog

PR #861, opened 2018-05-29
Bz#1400370 container entity content moved

dayleparker:
Hi Suyog,

I've moved the content/images now on container resources (and kept the term 'container entities' as it looks like it comes from the UI) from Managing Providers to the Managing Infrastructure guide.

I also saw some structural nesting issues with the Security Groups sections of the Managing Infrastructure guide so I fixed those as well.

As it turns out, the section "cross-providers insight" became a bit of an orphan after moving the content out of Managing Providers. I made it its own chapter at the end of the guide.

However, I don't think it's providing anything that useful, and I'd say it's well covered by Release Notes and the Managing Providers guide. What are your thoughts on just deleting this subsection?

Build previews to the two guides are in https://bugzilla.redhat.com/show_bug.cgi?id=1400370

Thanks for your review!

Cheers,
Dayle


suyogsainkar:
Hey Dayle, as discussed I'm merging this content. We can revisit the section on Cross-Providers Insight later. Cheers, Suyog

PR #862, opened 2018-05-29
Fix productname attributes in HBAC docs

hairmare:
Uses the proper `product-title_short_l` attribute in ipa hbac commands.


dayleparker:
Thanks for the update @hairmare! LGTM 👍

PR #901, opened 2018-07-06
Added the newly supported currencies in the Chargeback rates section.

suyogsainkar:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1590037


dayleparker:
Hi Suyog,

The changes look good, but there is a typo in "Norwegian". Since I also have a suggestion I didn't fix it myself....

I would suggest changing "Renminbi (CNY)" to "Chinese Yuan Renminbi (CNY)" to be consistent with other currencies (and more clear/findable), where we include the country name. I think GBP is the exception -- according to Wikipedia there are various different countries' pounds under that one name, so I think it's OK to leave as is to it encompasses the various pound types. https://en.wikipedia.org/wiki/Pound_sterling

What do you think?
Cheers,
Dayle

suyogsainkar:
@dayleparker Hi Dayle, many thanks for your review and comments. I've fixed the typo and applied your suggestion about CNY; also, left Pound sterling as is. It's ready for reviewing/merging again :) Cheers, Suyog

dayleparker:
Thanks for making those updates @suyogsainkar! Looks great to me :+1:

PR #902, opened 2018-07-09
Added a note about the ability to apply advanced settings.

suyogsainkar:
This PR addresses the following RFE bugs:
https://bugzilla.redhat.com/show_bug.cgi?id=1590042
https://bugzilla.redhat.com/show_bug.cgi?id=1590049

### DIFF
--- a/doc-General_Configuration/topics/Configuration.adoc
+++ b/doc-General_Configuration/topics/Configuration.adoc
@@ -1496,12 +1496,13 @@ image:use-configuration-settings.png[]
 [[advanced-settings]]
 ===== Advanced Settings
 
-You may be instructed by Red Hat to edit some configuration settings manually. This feature is available for a limited number of options and can only be used by users assigned the super administrator role. Changing
-settings using this procedure may disable your {product-title} server.
+You may be instructed by Red Hat to edit some configuration settings manually. This feature is available for a limited number of options and can only be used by users assigned the super administrator role. Changing settings using this procedure may disable your {product-title_short} server.
 
 [NOTE]
 =========
-Only make manual changes to your configuration files if directed to do so by Red Hat.
+* Only make manual changes to your configuration files if directed to do so by Red Hat.
+
+* {product-title_short} provides the ability to edit advanced settings for all appliances from the user interface. You can also specify advanced settings for entire zones, with the advanced settings applied to individual appliances in that zone overriding zone-level settings when set.
 =========
 
 [[editing-configuration-files-manually]]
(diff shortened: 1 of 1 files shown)


dayleparker:
Hi Suyog,

Reading the addition to the note, I think it makes sense to call out that feature where settings override entire-zone settings in the admonition. 

However, I think both of the new features could use a bit more explanation about how to use them. For example, which parameter in advanced settings would you change to set something for an entire zone? Maybe this would be good to mention in one of the parameter table in that section (perhaps in the "server" table? I'm not sure though....)

Also, I think it's really cool that you can change the advanced settings for other appliances from a totally different appliance. (Can any appliance be configured from any other appliance, or do roles matter?) I missed that fact at first with the wording of "all appliances", but noticed it when reading https://bugzilla.redhat.com/show_bug.cgi?id=1536524. I think this should be made a little more clear, and again I'd suggest an example of how this would look, or even where to set this in that really long advanced settings .yaml file.

 Also, I don't think this feature needs to be described in a note box, it could be explained in the same section, maybe after the note.

Sorry for the extra work... but I think this can be really useful stuff for users with a little more detail.
Please let me know if you want to have a chat about it.

Cheers,
Dayle

suyogsainkar:
@dayleparker Hi Dayle, thanks for your review and comments. After looking at the engineering bugs/PRs[1] again, I can confirm that this enhancement is about exposing the advanced tab for all servers in every zone to allow modification of advanced config settings at zone and individual server level from the UI. 

> For example, which parameter in advanced settings would you change to set something for an entire zone? 
> Also, I think it's really cool that you can change the advanced settings for other appliances from a totally different appliance. (Can any appliance be configured from any other appliance, or do roles matter?)

-The existing procedure to edit advanced settings/config files still remains almost the same and applies here for all appliances, including remote appliances. You can click the zone and the server under it for which you want to edit the settings, and then click the Advanced tab (now exposed for all servers). You can then modify the required parameter(s) for that zone - listed under 4.1.4.6.2. Configuration Parameters section / tables, so no particular parameter is really linked to this "new" feature. 
Also, "This feature is available for a limited number of options and can only be used by users assigned the super administrator role" - is already mentioned at the start of the section. 

I think this capability suits well as a note which is appropriately placed at the start of the section before the procedure to edit advanced config settings file and the configuration parameters' tables. I have made a small change in the note text and in one of the steps in the procedure. I have also changed the Note box to an Important box. 

I hope this helps, and hopefully makes things a bit more clear but happy to discuss further. Let me know what you think. 

Thanks,
Suyog

dayleparker:
Hi Suyog,

Sorry for my slow reply... this got lost along the way.
Thanks for the added detail -- I think maybe I was understanding the feature as something else. The edits you made look good -- i just removed a few words to make the last part of the sentence a little clearer when reading (hope that is ok). Good idea to put it in an important box, too.

Cheers,
Dayle

suyogsainkar:
@dayleparker No problem, Dayle - thanks for merging.

PR #937, opened 2018-08-13
Updated the repo list in the procedure for registering a CF appliance

suyogsainkar:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1612649


suyogsainkar:
@adahms Hi Andrew, I've created the following new attributes for the repo names with some variations for the way they can appear in docs (single line, multi-line with line breaks, and as separate entries) to the attributes file for both cfme and miq:
- :product-repo_list:
- :product-repo_list_lb:
- :product-repo_cfme:
- :product-repo_ansible:
- :product-repo_extras:
- :product-repo_optional:
- :product-repo_server:

I've updated the list of repositories where it appears in the General Configuration (one line) and High Availability Guide (separate entries) using the newly created attributes. The attribute with line breaks (:product-repo_list_lb) also works fine for the HA content, however I was unable to apply formatting in that case (i.e. * for unordered list) therefore decided to use individual entries there instead. Can you please take a look and let me know what you think. 
Thank you, 
Suyog

suyogsainkar:
@cbudz Hi Chris,
Would you mind taking a look at these updates? I have completed all of the tasks as part of the bug, but we can revisit the repo attributes based on our team's requirements and how we want to use them. Let me know what you think.
Thanks,
Suyog

cbudz:
Looks good, thanks Suyog. Merging this request.

PR #949, opened 2018-08-23
add reset vm feature

cbudz:
Hey Dayle,

Quick addition here for review. I've just added the 'reset' feature to the power state control list.

-Chris


dayleparker:
Looks good to me, thanks Chris!

PR #953, opened 2018-08-30
Note about icon appearance delay

apinnick:


adahms:

PR #964, opened 2018-09-18
Add note re:config settings to adding tower provider

cbudz:
This PR adds a note to Sec 3.2.2 informing the user that a configuration setting on the provider side must first be enabled before the Tower provider can be added to the CFME inventory.


cbudz:
Merging PR.

PR #986, opened 2018-10-12
PostgreSQL 9.4 is no longer supported, replace with 9.5

himdel:
(Not sure if ManageIQ 4.x still applies to current releases, never heard that numbering.)


cbudz:
Hi @himdel What version does this apply to for backporting in docs?

cbudz:
Thanks, @himdel. I'll get this backported to Gaprindashvili and Hammer.

PR #991, opened 2018-10-23
Bz#1640943

apinnick:


doc-Infrastructure_Migration_Solution_Guide/cfme/doc/modules/con_Prerequisites.adoc
```diff
@@ -32,7 +33,12 @@ The following ports must be enabled:
 For more information, see link:https://access.redhat.com/articles/417343[Ports used by Red Hat CloudForms Management Engine 5.1 and above].
 
 Data and ISO storage domains attached to the RHV data center::
-See link:https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.2/html-single/administration_guide/#sect-Preparing_and_Adding_NFS_Storage[Preparing and Adding NFS Storage].
+See link:https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.2/html-single/administration_guide/#chap-Storage[Storage] in the _Red Hat Virtualization Administration Guide_.
++
+[NOTE]
```
cbudz: Hi Atival. I would recommend breaking out the last phrase, "not local storage" into its own sentence to support readability. "Local storage is not supported for Red Hat Virtualization."

apinnick: If you think it's more readable as two sentences, I would change that to "Local storage is not supported." It's not supported for RHV or any other product and won't be because the production environments use shared storage.

cbudz:
Excellent, thanks for making that update Atival.

PR #1050, opened 2019-01-09
minor wording fix to include editing lines as well as adding

dayleparker:


dayleparker:
backporting PR 1048 and 1049

PR #1058, opened 2019-01-17
Bz#1578149 updatepods

dayleparker:
Added to Installing on OpenShift guide instructions for updating the appliance with a new container image, and instructions for updating the appliance by updating the template and image.

### DIFF
--- a/doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
+++ b/doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
@@ -150,7 +150,7 @@ After the restore job is complete, you can scale the pods back up.
 [[updating-pods]]
 === Updating {product-title_short}
 
-To apply package updates to a {product-title_short} appliance running on OpenShift Container Platform, delete the pods and redeploy them using updated {product-title_short} container images. To update your environment with a new template, follow the steps in xref:modified-template-update[].
+Apply package updates to your {product-title_short} appliance from OpenShift Container Platform by deleting the pods and redeploying them using updated {product-title_short} container images. To update your environment with a new template, follow the steps in xref:modified-template-update[].
 
 ifdef::cfme[]
 See the https://access.redhat.com/documentation/en-us/red_hat_cloudforms/4.7/html/release_notes/index#technical_notes[Technical Notes] in the _Release Notes_ for a list of updated {product-title_short} container images and errata.
@@ -162,7 +162,7 @@ endif::cfme[]
 
 To deploy new container images for {product-title_short}, delete the pods, update the images, then redeploy the updated pods:
 
-. Scale the application StatefulSets to 0 replicas:
+. Delete the pods by scaling the application StatefulSets to 0 replicas:
 +
 ----
 $ oc scale statefulset cloudforms --replicas=0
@@ -202,7 +202,7 @@ Be careful to back up secrets in a secure location.
 $ oc export secret cloudforms-secrets > my_secrets.yml
 ----
 +
-. Scale the application StatefulSets to 0 replicas:
+. Delete the pods by scaling the application StatefulSets to 0 replicas:
 +
 ----
 $ oc scale statefulset cloudforms --replicas=0
@@ -216,13 +216,13 @@ $ oc scale statefulset cloudforms-backend --replicas=0
 If you customized any parameters when originally deploying the application (parameters used with the `oc new-app` command in xref:deploying-the-appliance[]), you must set the same values in the `oc process` command here.
 ====
 +
-Use the following command to apply changes to the project if your environment uses a database stored in a pod within the cluster (the default configuration):
+* For environments using a database stored on a pod within the cluster (the default configuration), specify the {product-title_short} template:
 +
 ----
 $ oc process -p APPLICATION_REPLICA_COUNT=0 -l app=cloudforms,template=cloudforms -f cfme-template.yaml | oc replace -f -
 ----
 +
-If you are using a database external to the OpenShift cluster, specify the external database template:
+* For environments using a database external to the OpenShift cluster, specify the {product-title_short} external database template:
 +
 ----
 $ oc process -p APPLICATION_REPLICA_COUNT=0 -l app=cloudforms,template=cloudforms-ext-db -f cfme-template-ext-db.yaml | oc replace -f -
--- a/doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
+++ b/doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
@@ -145,6 +145,41 @@ $ oc create -f cfme-restore-job.yaml
 
 After the restore job is complete, you can scale the pods back up.
 
+
+
+[[updating-pods]]
+=== Updating {product-title_short} Pods
+
+To apply package updates to a {product-title_short} appliance running on OpenShift Container Platform, deploy the new {product-title_short} container image with the following steps:
+
+. Scale both StatefulSets to 0 replicas:
++
+----
+$ oc scale statefulset cloudforms --replicas=0
+$ oc scale statefulset cloudforms-backend --replicas=0
+----
++
+. Patch in the new version:
++
+----
+$ oc patch statefulset cloudforms -p '{"spec": {"template": {"spec": {"containers": [{"name": "cloudforms", "image": "registry.redhat.io/cloudforms46/cfme-openshift-app-ui:<new-version-tag>"}]}}}}'
+$ oc patch statefulset cloudforms-backend -p '{"spec": {"template": {"spec": {"containers": [{"name": "cloudforms", "image": "registry.redhat.io/cloudforms46/cfme-openshift-app:<new-version-tag>"}]}}}}'
+----
++
+. Scale the StatefulSets back to your previous values:
++
+----
+$ oc scale statefulset cloudforms --replicas=<old-value>
+$ oc scale statefulset cloudforms-backend --replicas=<old-value>
+----
+
+Your {product-title_short} pods are now updated.
+
+[NOTE]
+====
+See the https://access.redhat.com/documentation/en-us/red_hat_cloudforms/4.7/html/release_notes/index#technical_notes[Technical Notes] in the _Release Notes_ for a list of updated {product-title_short} container images and errata.
+====
+
 [[uninstalling]]
 === Uninstalling {product-title} from a Project
 
--- a/doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
+++ b/doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
@@ -148,37 +148,129 @@ After the restore job is complete, you can scale the pods back up.
 
 
 [[updating-pods]]
-=== Updating {product-title_short} Pods
+=== Updating {product-title_short}
 
-To apply package updates to a {product-title_short} appliance running on OpenShift Container Platform, deploy the new {product-title_short} container image with the following steps:
+To apply package updates to a {product-title_short} appliance running on OpenShift Container Platform, delete the pods and redeploy them using updated {product-title_short} container images. If you have modified the template, this requires additional steps in xref:modified-template-update[].
 
-. Scale both StatefulSets to 0 replicas:
+ifdef::cfme[]
+
+See the https://access.redhat.com/documentation/en-us/red_hat_cloudforms/4.7/html/release_notes/index#technical_notes[Technical Notes] in the _Release Notes_ for a list of updated {product-title_short} container images and errata.
+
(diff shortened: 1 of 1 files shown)


doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
```diff
@@ -145,6 +145,133 @@ $ oc create -f cfme-restore-job.yaml
...
+[IMPORTANT]
+====
+If you are updating a {product-title} 4.6.0 environment to version 4.6.1, follow the steps in xref:modified-template-update[]. When updating between these versions, the template must also be updated to reflect the updated ConfigMap mount path for `postgresql`. 
+====
+endif::cfme[]
+
+ifdef::miq[]
+[IMPORTANT]
+====
+If you are updating a {product-title} 5.9.0 environment to version 5.9.1, follow the steps in xref:modified-template-update[]. When updating between these versions, the template must also be updated to reflect the updated ConfigMap mount path for `postgresql`.
```
carbonin: 4.6 and 5.9 are both downstream versions. I don't think this applies to upstream. I believe manageiq had the templates updated for Gaprindashvili-1.

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
```diff
@@ -145,6 +145,133 @@ $ oc create -f cfme-restore-job.yaml
 
 After the restore job is complete, you can scale the pods back up.
 
+
+
+[[updating-pods]]
+=== Updating {product-title_short}
+
+To apply package updates to a {product-title_short} appliance running on OpenShift Container Platform, delete the pods and redeploy them using updated {product-title_short} container images. If you have modified the template, this requires additional steps in xref:modified-template-update[].
```
carbonin: This says "If **you** have modified the template" .. That seems a bit confusing. The condition where a user would need to deploy an updated template is when we release a template with changes. It doesn't really have anything to do with if the user made modifications.

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
```diff
@@ -145,6 +145,133 @@ $ oc create -f cfme-restore-job.yaml
...
+====
+If you are updating a {product-title} 5.9.0 environment to version 5.9.1, follow the steps in xref:modified-template-update[]. When updating between these versions, the template must also be updated to reflect the updated ConfigMap mount path for `postgresql`.
+====
+endif::miq[]
+
+
+[[image-update]]
+==== Updating the {product-title_short} Container Images
+
+To deploy new container images for {product-title_short}, delete the pods, update the images, then rebuild the updated pods:
```
carbonin: I don't think "rebuild" is the right term here. Maybe re-deploy?

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
```diff
@@ -145,6 +145,133 @@ $ oc create -f cfme-restore-job.yaml
...
+$ oc scale statefulset cloudforms-backend --replicas=<old-value>
+----
+
+Your {product-title_short} environment now updated.
+
+
+[[modified-template-update]]
+==== Updating the {product-title_short} Container Images and Template
+
+To update a {product-title_short} deployment with a modified template, update the container images and the template using the following steps:
```
carbonin: Similar to my comment at the top. It's not about if the current template is modified, but if you're trying to affect template changes. Does that make sense?

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
```diff
@@ -145,6 +145,133 @@ $ oc create -f cfme-restore-job.yaml
...
+----
+$ oc scale statefulset cloudforms --replicas=0
+$ oc scale statefulset cloudforms-backend --replicas=0
+----
++
+. Patch in the new version of the images:
++
+----
+$ oc patch statefulset cloudforms -p '{"spec": {"template": {"spec": {"containers": [{"name": "cloudforms", "image": "registry.redhat.io/cloudforms46/cfme-openshift-app-ui:<new-version-tag>"}]}}}}'
+$ oc patch statefulset cloudforms-backend -p '{"spec": {"template": {"spec": {"containers": [{"name": "cloudforms", "image": "registry.redhat.io/cloudforms46/cfme-openshift-app:<new-version-tag>"}]}}}}'
```
carbonin: This step is not necessary here because the new image version will be built into the new template.

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
```diff
@@ -145,6 +145,133 @@ $ oc create -f cfme-restore-job.yaml
...
+====
+* If you customized any parameters when originally deploying the application (parameters used with the `oc new-app` command in xref:deploying-the-appliance[]), you must set the same values in the `oc process` command here.
+
+ifdef::cfme[]
+* If you are updating a {product-title} 4.6.0 environment to version 4.6.1, update the `postgresql` ConfigMap mount path from `postgresql-config` to `postgresql-cfg`.
+endif::cfme[]
+
+ifdef::miq[]
+* If you are updating a {product-title} 5.9.0 environment to version 5.9.1, update the `postgresql` ConfigMap mount path from `postgresql-config` to `postgresql-cfg`.
+endif::miq[]
```
carbonin: The user doesn't need to do this manually, if they're using the new template it will contain the change.

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
```diff
@@ -145,6 +145,133 @@ $ oc create -f cfme-restore-job.yaml
...
+====
++
+----
+$ oc process -p APPLICATION_REPLICA_COUNT=0 -l app=cloudforms,template=cloudforms -f cfme-template.yaml | oc replace -f -
```
carbonin: This is specifically the template for users using the podified database (rather than one external to the openshift cluster), do we want to generalize the template file name here? All the other steps should be the same.

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
carbonin: Also if the user wants to use a version different from the default in the new template, they can set that using the parameters in the `oc process` command.

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
dayleparker: Ah ok. I've deleted this note entirely now as it doesn't seem necessary to mention it as the template update will resolve this automatically.

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
dayleparker: Fixed.

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
dayleparker: I've updated the wording, hopefully it makes more sense now.

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
dayleparker: Makes sense -- I had confused this idea with the custom parameters before I think.

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
dayleparker: I've deleted this and the previous note about this now, thanks for clarifying that.

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
dayleparker: Ah cool - I've deleted this step now.

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
dayleparker: Good catch -- I've updated this step, and the command in two places (template name, and .yaml file) to match what's documented in the section  "1.3.1. Deploying the CloudForms Appliance Using an External Database". Does that look right?

dayleparker:
Hi @carbonin,
Thanks for all your comments here, much appreciated. 

The last thing I'm still a bit confused about is the relationship between the templates and images. IIUC, when we release new container images, Red Hat also updates and releases a new template that points to the new images, correct?

If I'm understanding that correctly, do you think there's a use case (for downstream users) for the first procedure on just updating just the container images?

Alternatively, I could move the container image update part to another section in the guide, if it serves a purpose other than applying updates from errata. Curious about your thoughts. Thanks so much!

dayleparker:
Hi Chris,

Would you mind doing the final review on this content please?
Let me know if you have any thoughts or you think it fits OK in this part of the guide (I have yet to hear if an upgrade part for major releases should be added).


Thanks for your review :)

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
```diff
@@ -145,6 +145,112 @@ $ oc create -f cfme-restore-job.yaml
...
+See the https://access.redhat.com/documentation/en-us/red_hat_cloudforms/4.7/html/release_notes/index#technical_notes[Technical Notes] in the _Release Notes_ for a list of updated {product-title_short} container images and errata.
+endif::cfme[]
+
+
+[[image-update]]
+==== Updating the {product-title_short} Container Images
+
+To deploy new container images for {product-title_short}, delete the pods, update the images, then redeploy the updated pods:
+
+. Scale the application StatefulSets to 0 replicas:
```
cbudz: Hey Dayle,

Is this command run on the OCP server? Do we need a command instructing users to SSH to te OCP server first, just to foolproof things?

cbudz: Also should StatefulSets have any markup around it if its a config file parameter? (I'm assuming that's what it is)

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
```diff
@@ -145,6 +145,112 @@ $ oc create -f cfme-restore-job.yaml
...
+----
++
+. Apply the changes to the project, specifying the template appropriate to your environment's database configuration.
++
+[IMPORTANT]
+====
+If you customized any parameters when originally deploying the application (parameters used with the `oc new-app` command in xref:deploying-the-appliance[]), you must set the same values in the `oc process` command here.
+====
++
+Use the following command to apply changes to the project if your environment uses a database stored in a pod within the cluster (the default configuration):
```
cbudz: Here do think there's value in identifying the database type first?

Ex.

1. For Environments using databases stored within a cluster:
 
Do x

2. Databases external to the OCP Cluster:

Do Y

This came up in a recent minimalism session, hence applying it here.

dayleparker: Excellent idea -- that is a much better/straightforward way than what I'd written before. Let me know what you think of the new wording.

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
dayleparker: Hmm, good points here... it's true in other docs we might start each procedure with the SSH command (or mention it) for completeness. However, everything in this guide (except the last step of 'Logging into CloudForms") is done from OpenShift to administer CloudForms, and I believe OCP admins and regular users should know to run the "oc" commands from the OpenShift server. 

I'm really not sure whether we should mark up StatefulSets (it's an OpenShift object --> a set of pods). We haven't in a previous section, and I wonder if marking it up creates a situation where we need to mark up all the OpenShift terms, like template, pods, etc? (the capitalization makes it look weirder than the other terms)

I lean towards leaving the above two things as is, but what do you think?

dayleparker:
Hey Chris,
Thanks for your comments and review! I've made a few edits -- let me know if you think things are a little more clear, and what your thoughts are on the SSH & StatefulSets topics are.
Cheers, Dayle

doc-Installing_on_OpenShift_Container_Platform/topics/post-installation.adoc
cbudz: I think those explanations clear up any questions I had. The capitalization of StatefulSets did make me immediate think it was a config parameter, hence asking about markup. Thanks!

PR #1074, opened 2019-02-12
Bz#1513042 centraladmin

dayleparker:
Hi Suyog,
Can you please review these updates for https://bugzilla.redhat.com/show_bug.cgi?id=1513042 and let me know if you think it's all clear? Feedback/edits welcome :)
Thank you,
Dayle


suyogsainkar:
Thanks Dayle! It seems not a lot has changed since the feature was introduced but these updates look good. The Self Service UI does not have parity with the classic UI for several features but it makes sense to call out that central admin isn't supported at all in the SSUI. Merging to master.

dayleparker:
Thanks for reviewing and merging this, Suyog!

PR #1087, opened 2019-03-12
1.1 fix

apinnick:
Fixed heading level and other updates


cbudz:
I've merged this updates. @apinnick

PR #1131, opened 2019-05-08
Add note and procedure for wal keep segments config

cbudz:


suyogsainkar:
Thanks, Chris. Looks good to me; merging to master.

PR #1161, opened 2019-06-20
Added navigation links

apinnick:


cbudz:
ok, changes look good.

PR #1170, opened 2019-07-02
Fix template path for saml config

apagac:
Path for the templates has changed.

### DIFF
--- a/auth/saml.adoc
+++ b/auth/saml.adoc
@@ -44,7 +44,7 @@ First ssh to the appliance as root, then create that directory:
 Copy the Apache remote user and SAML template configuration files:
 
 ```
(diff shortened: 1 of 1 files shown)


jvlcek:
@apagac The old path is for ManageIQ and the change you made is for downstream.
Don't we want to keep these docs accurate for upstream, ManageIQ?

JoeV

apagac:
@jvlcek ah, I didn't realise that. You are most probably right.

jvlcek:
@apagac Thank you.

The paths for the templates can be seen in the source code for the miqldap_to_sssd conversion tool here:
https://github.com/ManageIQ/manageiq/blob/master/tools/miqldap_to_sssd/auth_template_files.rb#L8

PR #1172, opened 2019-07-04
Ims 1.2

apinnick:
Doc preview: http://file.tlv.redhat.com/~apinnick/IMS_1.2

### DIFF
--- a/doc-Infrastructure_Migration_Solution_Guide/cfme/IMS_1.1/modules/proc_Installing_rhv_4_2.adoc
+++ b/doc-Infrastructure_Migration_Solution_Guide/cfme/IMS_1.1/modules/proc_Installing_rhv_4_2.adoc
@@ -21,11 +21,13 @@ For details, see https://access.redhat.com/articles/417343[Ports used by Red Hat
 [NOTE]
 ====
 IMS only supports shared storage, such as NFS, iSCSI, or FCP. Local storage is not supported.
+
+Although the ISO storage domain has been deprecated, it is required for migration.
 ====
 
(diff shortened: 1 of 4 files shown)


doc-Infrastructure_Migration_Solution_Guide/cfme/IMS_1.1/modules/proc_Installing_rhv_4_2.adoc
```diff
@@ -21,11 +21,13 @@ For details, see https://access.redhat.com/articles/417343[Ports used by Red Hat
 [NOTE]
 ====
 IMS only supports shared storage, such as NFS, iSCSI, or FCP. Local storage is not supported.
+
+Although the ISO storage domain has been deprecated, it is required for migration.
```
cbudz: My only concern here is the link through below to the RHV topic, where it states that ISO storage domain is deprecated, proceeds to instruct users to upload ISO images to a data domain, then reverts back to using ISO storage domain as the location. It gets a little confusing, but if the admins performing theses tasks are knowlegable on the concepts I guess it shouldn't cause much trouble

cbudz:
@apinnick I've just left one comment, but am merging this, since my concern is on the RHV docs side.

doc-Infrastructure_Migration_Solution_Guide/cfme/IMS_1.1/modules/proc_Installing_rhv_4_2.adoc
apinnick: ISO domain is deprecated for RHV but required for IMS. That's why I added the note. As long as RHV provides documentation for adding ISO domains and uploading to ISO domains, I'll have to link to them. When RHV stops supporting this, the documentation will have to be moved to IMS.

PR #1173, opened 2019-07-08
Clarified adding OSP as cloud provider and updated link, OSP V2V imag…

apinnick:
…e versions for 1.1 and 1.2


cbudz:
Looks good, @apinnick

PR #1178, opened 2019-07-14
Multiple conversion hosts for load-balancing

apinnick:


cbudz:
Looks good, merging this PR @apinnick

PR #1205, opened 2019-08-08
[HAMMER] Correct doc based on DDF feedback

cbudz:


suyogsainkar:
@cbudz Looks good to me; merging. Thanks.

PR #1206, opened 2019-08-08
[IVANCHUK] Adding correction for name attribute

cbudz:


suyogsainkar:
@cbudz Looks good to me; merging. Thanks.

PR #1209, opened 2019-08-16
[FINE] Remove extra brace on fine version of add_resource

d-m-u:
I know, it doesn't matter cause it's old, but it's still wrong and I wanted it to be right.


abellotti:
Nice find @d-m-u thanks for fixing this. 👍

PR #1239, opened 2019-09-04
fixed file name

apinnick:
@cbudz  I just noticed that I forgot to use the file-naming convention I've been following for manageiq (too much time in openshift docs....). I've updated the file name and checked that the docs build. Please merge. Thanks!


cbudz:
no problem, change looks good. Merging

PR #1241, opened 2019-09-06
Update Ansible Method and add link to export datastore classes

cbudz:
Hi Dayle,

Just a quick update here to satisfy this BZ. I've updated the Ansible Method section of documentation to include the Vault field, and added some text at the bottom informing users that this method can be exported with some reasons and a link to the relevant section. 

### DIFF
--- a/doc-Scripting_Actions/topics/Ansible_method.adoc
+++ b/doc-Scripting_Actions/topics/Ansible_method.adoc
@@ -23,6 +23,7 @@ image:../images/1862.png[image] (*Add a New Method*).
 . Select a playbook *Repository* from the list.
 .. Choose a *Playbook* to use.
 .. Select the *Machine Credential* the playbook will use when it runs.
+.. Select the *Vault Credential* to use.
 .. From the *Cloud Type* list, select a cloud provider.
 .. Choose the *Cloud Credential* that corresponds to the selected cloud type.
 . Specify the *Hosts* on which the playbook will run. Choose *Localhost* or provide unique values in the *Specify host values* field.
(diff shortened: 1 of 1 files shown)


dayleparker:
Hi Chris,
The update looks good to me -- however I am definitely not an expert on the Automate or Ansible parts of CloudForms, so I reviewed this with extra care (maybe too much?) that led to a couple (possibly n00b) questions. I agree (reading the original PR (https://github.com/ManageIQ/manageiq-automation_engine/pull/303) this does seem like more of a backend change that doesn't need a lot of documentation, other than a release note, and a bit of text about exporting a domain with an Ansible method, as you added.

To be thorough, I ran through the procedure on the enablement appliance, and everything works great. I assume that "Vault credentials" are familiar to any Ansible admin and they would know what to enter in this field (right?). A quick Google to Ansible's docs make it all pretty clear if more info is needed.

One correction: 
The only small thing I can see that needs correcting in the section (in the text already there, not from your commit) is in step 5, "Playbook" is capitalized, and the step could be more explicit as 'Select "Playbook" from the "Type" dropdown menu" (or similar, rather than 'Main info").

Re: the text at the bottom:
* Placement of the note and link are great -- to me this links to the best place in the guide, and I tried out the export procedure just for fun, and it was also very easy to follow. 
* Terminology -- The only thing I'm still unsure about is the language -- ie. the original BZ and eng PR talks about exporting a 'domain' (and yes, when I hit the Export button and unzipped the file, I easily found a YAML file defined as a domain), while the note, and the section in the docs talks about exporting a class. Should this be clarified a little, or is it obvious that exporting a class also exports the domain (which I think contains the classes and methods)? Do you think it would be worth it to mention somewhere --in the note in the Ansible section, or in the Export/Import Class section -- about the ability to export a **domain**, in case a customer is searching for that term specifically? I am just not sure how Automate users think...

Apologies for the lengthy explanation -- just wanted to try and best explain my process in case you're looking for something different in the review. Let me know what you think :)

cbudz:
Hey @dayleparker thanks for the feedback. I've updated the file the reflect suggested changes. Please feel free to merge if everything looks good.

dayleparker:
Looks great to me!

PR #1249, opened 2019-09-13
edited tenant mapping content to mention user sync

dayleparker:
Hi Chris,

Would you mind giving this a quick review to be sure I've satisfied the BZ request, but also haven't made it confusing with my changes?

I think what we already had in the Managing Providers guide was OK, but I have added a mention about user syncing it to make it more clear that enabling tenant mapping includes this. I also reworded the step to enable tenant mapping a little to provide the action first, then the additional info so it's easier to skim.

(I really liked the explanation in https://github.com/ManageIQ/manageiq-ui-classic/pull/2495 but couldn't figure out how to incorporate it)

I've attached a PDF of the whole guide to the BZ for easier review.
Any feedback welcome. 
Thanks, Dayle


cbudz:
Hey Dayle, I think this reads well and the explanations are clear. Merging.

PR #1259, opened 2019-09-18
Rhv host subscriptions

apinnick:
@cbudz You can ignore most of the changes because I was cleaning out cache. I updated prerequisites for RHV conversion hosts. If system has been active for a while, the host need to have valid subscriptions.

Doc previews:

http://file.tlv.redhat.com/~apinnick/IMS_1.1/#Configuring_the_1_1_rhv_conversion_hosts_for_vddk
http://file.tlv.redhat.com/~apinnick/IMS_1.2/#Configuring_the_1_2_rhv_conversion_hosts_for_vddk


cbudz:
@apinnick should we provide more detail on how to ensure a valid subscription? Is this through Red Hat Subscription Manager or another subscription system? Is there any documentation we can link to?

apinnick:
> @apinnick should we provide more detail on how to ensure a valid subscription? Is this through Red Hat Subscription Manager or another subscription system? Is there any documentation we can link to?

Well, here's the thing. RHV now has 4 different installation guides (SHE with UI, SHE from CLI, Standalone engine with local DBs, Standalone engine with remote DBs) x 2 host possibilities (RHEL, RHVH), which means I would require 8 links. So I think it's better not to include a link.

Yes, it's RH Subscription Manager. I'll add this.

cbudz:
lgtm, merged.

PR #1275, opened 2019-09-30
added note about having a matching playbook name on the target

dayleparker:


dayleparker:
Christian Jung has reviewed the text in https://bugzilla.redhat.com/show_bug.cgi?id=1750784; merging.

PR #1277, opened 2019-10-01
added directories to back up

dayleparker:
These directories weren't mentioned in the docs, but otherwise we have instructions for backing up and restoring the database. I've added this to the procedure.


doc-General_Configuration/topics/Configuration.adoc
```diff
@@ -2913,6 +2913,11 @@ You can reset the replication relationship between the global copy and remote co
 
 {product-title} supports database backups to *Network File System (NFS)*, *Samba*, *Amazon Web Service(AWS) S3* and *OpenStack Swift* storage. 
 
+This procedure backs up the following directories:
```
carbonin: These are actually all files, not directories. And also this process here only backs up the database, not files from the filesystem. The user would need to keep these somewhere else to ensure they can be restored.

Fryguy:
Closing as master has been completely changed to Markdown.  If this is still a feature, please reopen.

PR #1283, opened 2019-10-08
Bug BZ1647742 - Removed item numbers

jhoyt-rh:
For Bug BZ1647742 -- removed menu item numbers. Number in manual was incorrect because number had changed. We are removing the numbers since numbers will continue to change going forward as new items are added. Now the manual just instructs the user to select the action, rather than a numbered action.


cbudz:
lgtm 👍

PR #1284, opened 2019-10-08
Add section on passing variables

cbudz:


dayleparker:
Hey Chris,
This reads well and has all the info I'd think an automate designer would need to do this.
I'm definitely not an Ansible expert but what I'm understanding here is wherever you use "pb_var1a" or "pb_var1b", in any successive playbook, you get whatever is inside those brackets, and it made sense with looking up the set_stats Ansible docs. +1 to the excellent Flintstones Easter egg :) (Let me know if you're looking for other feedback)

Would you mind adding a full stop to the last sentence? Then I think it's good to merge.

cbudz:
@dayleparker thanks I've added that final stop.

PR #1288, opened 2019-10-16
Fix formatting of lists: add an empty line before them.

martinpovolny:
fixes for:
  * http://www.manageiq.org/docs/reference/latest/api/overview/http
  * http://www.manageiq.org/docs/reference/latest/api/overview/json#actions


abellotti:
Thanks @martinpovolny for finding and fixing this 👍

PR #1289, opened 2019-10-30
Add default cfme credentials for initial login

cbudz:
Hey Dayle,

Quick addition here re:DDF request.

 - chris


dayleparker:
Looks great! Thanks for making this change so quickly :)

PR #1290, opened 2019-11-06
[IVANCHUK] Test ruby 2.5.5, see: https://github.com/ManageIQ/manageiq/pull/19468

d-m-u:
Test ruby 2.5.5, see: https://github.com/ManageIQ/manageiq/pull/19468

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: ruby
 rvm:
-- 2.4.5
-- 2.3.8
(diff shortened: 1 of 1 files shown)


Fryguy:
Technically, the branch versions don't matter because the way builds work they are all from master.

Separately, this is more of a website-build-system question and what version of Ruby is available there.  @bdunne @cybette ?

PR #1292, opened 2019-11-15
Enhancing the SAML authentication page

abellotti:
Enhancing the SAML authentication page to talk about the new options that allows unattended SAML configuration.

### DIFF
--- a/auth/saml.adoc
+++ b/auth/saml.adoc
@@ -18,6 +18,7 @@ The current implementation only secures the Appliance's Web administrative UI wi
 2.  <<configuring-saml, Configuring SAML>>
 3.  <<configuring-admin-ui, Configuring the Administrative UI>>
 4.  <<re-enabling-local-login, Re-Enabling Local Login>>
+5.  <<configuring-saml-via-cli, Configuring SAML using the CLI>>
 
 [[requirements]]
 == Requirements
@@ -27,7 +28,7 @@ The following is needed in order to enable SAML Authentication to the Appliance:
 * A CentOS/RHEL 7.2 based Appliance
 * A SAML Identity Provider, (e.g. KeyCloak 1.8 or Later)
 
-[[configure-saml]]
+[[configuring-saml]]
 == Configure SAML 
 
 All SAML related certificates and keys are accessed from /etc/httpd/saml2/
@@ -171,7 +172,7 @@ Mappers.
 *Note:* The fullName attribute was not available in the default database as of this writing
 and must be added to each user as a user attribute.
 
-[[configure-admin-ui]]
+[[configuring-admin-ui]]
 == Configure Administrative UI 
 
 After having configured Apache for SAML, the next step is to update the Appliance Administrative UI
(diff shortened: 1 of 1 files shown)


abellotti:
Keeping as WIP until the code is merged.

auth/saml.adoc
```diff
@@ -226,3 +227,67 @@ with enough administrative privileges to update it:
...
+                                      [--saml-enable-sso]
+```
+
+When --saml-client-host is not specified, the configured appliance host
+is used for creating the SP metadata. This host fqdn must be reachable from
+the SAML IDP.
+
+Configuring SAML requires the IDP metadata file, the --saml-idp-metadata option allows
+the user to specify a copy by file path that was downloaded to the appliance or by URL
+in which case the CLI downloads the copy from the IDP.
```
jvlcek: This wording seems odd to me. May I suggest you consider:

Configuring SAML requires the IDP metadata file. The --saml-idp-metadata option allows
the user to specify the file path of a copy that was downloaded to the appliance or by a URL,
in which case the CLI downloads the copy from the IDP.

auth/saml.adoc
```diff
@@ -226,3 +227,67 @@ with enough administrative privileges to update it:
...
+
+```
+--saml-idp-metadata=http://<idp-host:port>/auth/realms/<miq-realm>/protocol/saml/descriptor
+```
+
+
+In both cases, the IDP metadata file is copied to `/etc/httpd/saml2/idp-metadata.xml`
+
+By default, SSO is not enabled, so from the Appliance login page, the user clicks
+on Login In to Corporate System to get redirected to the SAML login page. With this option
```
jvlcek: Update the wording for the SAML login page is: 

From:

> Login In to Corporate  System

To:
> Log In to Corporate System

auth/saml.adoc
```diff
@@ -226,3 +227,67 @@ with enough administrative privileges to update it:
...
+To unconfigure the appliance from SAML Authentication and revert to Database authentication:
+
+```
+# appliance_console_cli --saml-unconfig
+```
+
+This will remove the Apache external authentication SAML configuration files and revert
+the appliance's authentication settings to Database mode.
+
+With both --saml-config and --saml-unconfig subcommands, the httpd service will be restarted
```
jvlcek: This functionality seems odd to me. Maybe an additional flag should be considered instead of using the two opposing flags to restart httpd, perhaps: `--restart_httpd`

abellotti: well, this has to be done regardless for the changes to take effect.  We do that for configuring ext-auth (ipa client/sssd/apache).  We could just drop it from the doc.

jvlcek: Got it. OK with me.

abellotti:
Hi @jvlcek I've unwipped this since the code is now merged, can you merge this ?  Thanks.

jvlcek:
@Fryguy beat me to it. Thank you @Fryguy

PR #1340, opened 2020-02-07
Remove timezone and date and time

jhoyt-rh:
Removing timezone and date and time, since they are not working when set from the Appliance Console. They used to work, so this is a regression.


cbudz:
All changes look good, requested content has been removed

PR #1347, opened 2020-02-20
Use updated retirement calls for service retirement 

d-m-u:
Pretty sure that this call will 💣 on anything recent.

I'm pretty sure there are more of these to fix but it has to be in follow-up

### DIFF
--- a/api/examples/retire_service.adoc
+++ b/api/examples/retire_service.adoc
@@ -11,7 +11,7 @@ POST /api/services
 [source,json]
 ----
 {
-  "action" : "retire",
+  "action" : "request_retire",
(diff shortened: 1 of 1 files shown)


suyogsainkar:
Thanks @d-m-u . Does this apply to I & H versions?
cc: @jhoyt-rh can you cherry-pick once confirmed / merged.

d-m-u:
@suyogsainkar when you were here, we discussed how these changes went into everything past the gap release

suyogsainkar:
@d-m-u oh, right. thanks for the reminder. merging now.

jhoyt-rh:
@suyogsainkar SO, this is to be cherry-picked to Ivanchuk and Hammer, right?

PR #1348, opened 2020-02-20
Update retire_services.adoc multiple service retirement call

d-m-u:
`retire` is the old call and will break, `request_retire` is what should be here
see https://github.com/ManageIQ/manageiq_docs/pull/1347 and http://talk.manageiq.org/t/retirement-as-a-request/4797


suyogsainkar:
LGTM. thanks.

PR #1355, opened 2020-02-25
[IVANCHUK] use updated retirement call

jhoyt-rh:
backported from PR #1347

Cherry-pick to Ivanchuk


suyogsainkar:
@jhoyt-rh looks good to me.

PR #1357, opened 2020-02-26
Use updated retire request for services

d-m-u:
Pretty sure we shouldn't be using the old retirement call here
@tinaafitz please review


d-m-u:
@suyogsainkar please review

suyogsainkar:
@d-m-u looks good to me; merging.

PR #1373, opened 2020-03-23
Update topic yaml for rest api

apinnick:


apinnick:
@Fryguy I'm not sure why I got a build error for that commit. Could you please take a look?

PR #1377, opened 2020-03-24
Ma manifest update

apinnick:


apinnick:
@RichardHoch Please review

Issue #1368, opened 2020-03-21
you can't Pausing / Resuming Containers Providers [sic] anymore


d-m-u:
Per https://github.com/ManageIQ/manageiq/pull/19915 I don't think this section should still be in our docs.

d-m-u:
Uhm, I think https://github.com/ManageIQ/manageiq_docs/blob/master/doc-Managing_Providers/topics/Pause_Resume_Containers_Providers.adoc needs to be changed a little then

PR #1404, opened 2020-04-07
Remove pause/resume documentation for Container Providers

chessbyte:
Fixes https://github.com/ManageIQ/manageiq_docs/issues/1368

it seems to have been removed in https://github.com/ManageIQ/manageiq/pull/19915 and https://github.com/ManageIQ/manageiq-api/pull/740

### DIFF
--- a/doc-Managing_Providers/topics/Containers_Providers.adoc
+++ b/doc-Managing_Providers/topics/Containers_Providers.adoc
@@ -83,10 +83,6 @@ include::Removing_Containers_Providers.adoc[]
 
 
 :leveloffset: 2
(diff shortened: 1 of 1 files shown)


Fryguy:
gah I merged, but realized after that you didn't remove the actual .adoc file...just the include link to it.

chessbyte:
@Fryguy I did not remove the .adoc file as I think we will want it when we actually get this functionality working.  But can remove if you think it just adds clutter and revisit when this feature is implemented properly.

Fryguy:
No, leaving it is fine...I just have to account for that in the markdown conversion, as the current script I have prunes things.

PR #1405, opened 2020-04-08
update retirement calls

d-m-u:
use retirement as a request calls instead of the older retirement calls

### DIFF
--- a/api/overview/json.adoc
+++ b/api/overview/json.adoc
@@ -171,7 +171,7 @@ POST /api/services
 [source,json]
 ----
 {
-  "action": "retire",
+  "action": "request_retire",
   "resources" : [
     { "href" : "http://localhost:3000/api/services/101" },
     { "href" : "http://localhost:3000/api/services/102",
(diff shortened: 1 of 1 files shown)


abellotti:
The API currently supports both, retire and request_retire.  Documentation should reflect this.

d-m-u:
Then we should make it clearer in these docs what the differences between the two are because I keep fixing bugs where people are using the wrong call.

abellotti:
That's fine, unless we are going to throw away/deprecate the old one, let's document both and clearly describe the difference.

d-m-u:
Alright but I would have thought that for this overview page especially we'd at least have examples that are up-to-date. Since the only reason we still have the old call is backwards compatibility and everything else is using the request call, it isn't really user-friendly to be using the old example here.

abellotti:
Ok, if the old one is kept around for compatibility, then we could deprecate it at some point.  Then, yes, let's just document request_retire. 👍

d-m-u:
Deprecating it would break retirement for any pre-Hammer state machine, which is why I think we don't want to deprecate it. But since its functionality has that limit, I do think these docs for any relevant release should be using the request call.

PR #1407, opened 2020-04-08
update final retirement call

d-m-u:
@abellotti could you please 👀 
thanks

### DIFF
--- a/api/reference/vms.adoc
+++ b/api/reference/vms.adoc
@@ -11,7 +11,7 @@ Management of virtual machines adds support of the following actions:
 | stop | Stops a virtual machine
 | suspend | Suspends a virtual machine
 | refresh | Refreshes a virtual machine
-| retire  | Retires a virtual machine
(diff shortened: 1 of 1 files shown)


api/reference/vms.adoc
```diff
@@ -163,6 +163,7 @@ GET /api/vms/:id?expand=accounts,software
   "action" : "request_retire"
 }
 ----
+Note that there are two retirement endpoints. Retirement as a request (RaaR) was added in our release of Hammer and replaces retire_now as the default retirement method. The retire_now method is kept for backwards compatibility but can be used only with a pre-Hammer Automate state machine.
```
abellotti, line 166: I wouldn't add this to the doc since we only want folks to use request_retire. Just the change to the table above to reflect "request_retire". Thanks.

abellotti:
- line 14 in that file (the table), the action name should change from "retire" to "request_retire"

Issue #238, opened 2017-02-24
outdated documentation for Creating Chargeback Rates


mkoura:
http://manageiq.org/docs/reference/latest/api/reference/chargebacks_rates#creating-rates
The request described in doc fails with
```
{
  "error": {
    "kind": "bad_request",
    "message": "Chargeable field can't be blank",
    "klass": "Api::BadRequestError"
  }
}
```
In current version also "chargeable_field_id" needs to be specified.

chessbyte:
@abellotti Is this old issue still valid?

abellotti:
yep, still an issue.

PR #1411, opened 2020-04-09
Fix example for POST /api/rates

abellotti:
Fix example for POST /api/rates
- was missing the now required chargeable_field_id.

/cc @lpichler please review 🙏

Fixes: https://github.com/ManageIQ/manageiq_docs/issues/238


chessbyte:
@lpichler please review

PR #1412, opened 2020-04-09
Add example dialog id to custom button create

d-m-u:
Otherwise I don't think we have an example showing that this is via the resource action. It is relevant all the way back to five ten and would be cool if it were in those docs.


d-m-u:
Found when looking at https://github.com/ManageIQ/manageiq-api/pull/796 
@miq-bot add_label jansa/yes, ivanchuk/yes, hammer/yes

d-m-u:
hey @abellotti any chance I could get your eyes on this sometime please?

PR #1414, opened 2020-04-16
Added synchronization to CF config

apinnick:


istein1:
Content looks good to me

Issue #1271, opened 2019-09-26
"Obtaining the Appliance" section needs updates


simaishi:
User Reference -> Installation -> Microsoft Azure (or Openstack/RHV), section 1.1

1. "navigate to [manageiq.org/download](http://manageiq.org/docs/reference/latest/doc-Installing_on_Microsoft_Azure/miq/manageiq.org/download)" link is broken.

1. step 2 to 4 talk about choosing platform/release from list and follow instruction, but we no longer use dropdown lists in the download page.

chessbyte:
@simaishi The broken links are fixed with this [PR](https://github.com/ManageIQ/manageiq_docs/pull/1402).

Any suggestions for new wording on steps 2-4?

simaishi:
I think the download page is simple enough that we can probably just say download an appropriate image?

PR #1430, opened 2020-04-23
Instructions did not match the updated download interface

davidbaren:
Fixes https://github.com/ManageIQ/manageiq_docs/issues/1271

### DIFF
--- a/doc-Installing_on_Microsoft_Azure/topics/installation.adoc
+++ b/doc-Installing_on_Microsoft_Azure/topics/installation.adoc
@@ -15,8 +15,7 @@ After you have completed all the procedures in this guide, you will have a worki
 ifdef::miq[]
 . In a browser, navigate to link:https://www.manageiq.org/download[].
 . Find the table with the relevant type of release.
-. Select *Microsoft Azure* from the list in the table.
-. Follow the instructions to complete the download of the appliance.
+. Click *Microsoft Azure* from the list to download the appliance image.
 endif::miq[]
 
 ifdef::cfme[]
--- a/doc-Installing_on_Microsoft_Azure/topics/installation.adoc
+++ b/doc-Installing_on_Microsoft_Azure/topics/installation.adoc
@@ -14,9 +14,9 @@ After you have completed all the procedures in this guide, you will have a worki
 
 ifdef::miq[]
 . In a browser, navigate to link:https://www.manageiq.org/download[].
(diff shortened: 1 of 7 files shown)


doc-Installing_on_Microsoft_Azure/topics/installation.adoc
```diff
@@ -14,9 +14,9 @@ After you have completed all the procedures in this guide, you will have a worki
 
 ifdef::miq[]
 . In a browser, navigate to link:https://www.manageiq.org/download[].
-. Select *Microsoft Azure* from the *--Choose your platform--* list.
-. Select *Stable* from the *--Choose a release--* list.
-. Follow the instructions to download the appliance.
+. Find the table with the relevant type of release.
+. Select *Microsoft Azure* from the list in the table.
+. Follow the instructions to complete the download of the appliance.
```
simaishi, line 19: Clicking the name will download the image and there is no further instruction, so I think those 2 can be combined to something like:

"Click Microsoft Azure from the list to download the appliance"

chessbyte:
@davidbaren thank you for making this PR.

I think we need similar changes for [Amazon Web Services](https://www.manageiq.org/docs/reference/latest/doc-Installing_on_AWS/miq/), [Google Compute Engine](https://www.manageiq.org/docs/reference/latest/doc-Installing_on_Google_Compute_Engine/miq/), [VMware vSphere](https://www.manageiq.org/docs/reference/latest/doc-Installing_on_VMware_vSphere/miq/), [OpenStack](https://www.manageiq.org/docs/reference/latest/doc-Installing_on_Red_Hat_Enterprise_Linux_OpenStack_Platform/miq/), [RHV](https://www.manageiq.org/docs/reference/latest/doc-Installing_on_Red_Hat_Virtualization/miq/), and [Microsoft SCVMM](https://www.manageiq.org/docs/reference/latest/doc-Installing_on_SCVMM/miq/).

You can either make these changes in this PR or in separate follow-up PRs.

Additionally, they probably should all say the same thing, so perhaps a new file should be created in the [common directory](https://github.com/ManageIQ/manageiq_docs/tree/master/common) that is leveraged by all of the above installation guides.

doc-Installing_on_Microsoft_Azure/topics/installation.adoc
davidbaren: I made the changes you suggested to Microsoft Azure and all of the other platforms as well.

PR #1437, opened 2020-04-29
Update tagging.adoc

d-m-u:
add lans to collections that support tagging since they do now: https://github.com/ManageIQ/manageiq-api/pull/804

closes https://github.com/ManageIQ/manageiq-api/issues/801


abellotti:
LGTM. Thanks @d-m-u for crossing the T's and dotting the I's. 👍

PR #1460, opened 2020-06-18
Add OIDC support for the Self-Service UI 

abellotti:
- Updated the OpenID-Connect authentication page to add support  the Self-Service UI.

### DIFF
--- a/auth/openid_connect.md
+++ b/auth/openid_connect.md
@@ -15,7 +15,7 @@ mod\_auth\_openidc module and should work with other OIDC Identity
 Providers.
 
 The current implementation secures the Appliance’s Web
-administrative and self service UI with OIDC.
+administrative and Self-Service UI with OIDC.
 
 1.  [Requirements](#requirements)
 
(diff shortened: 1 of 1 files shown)


auth/openid_connect.md
```diff
@@ -14,10 +14,8 @@ been tested with KeyCloak but is implemented generically using Apache’s
 mod\_auth\_openidc module and should work with other OIDC Identity
 Providers.
 
-The current implementation only secures the Appliance’s Web
-administrative UI with OIDC.
-
-**Note:** The Self-Service UI does not currently support OIDC.
+The current implementation secures the Appliance’s Web
+administrative and self service UI with OIDC.
```
jvlcek, line 18: @abellotti I think `self service` should be `Self-Service`. So this should read:
`administrative and Self-Service UI with OIDC.`

auth/openid_connect.md
```diff
@@ -65,6 +63,7 @@ following:
 | Client Protocol                    | openid-connect                                    |
 | Access Type                        | confidential                                      |
 | Valid Redirect URIs                | <https://<miq-appliance>>                         |
+|                                    | <https://<miq-appliance>/ui/service/>             |
```
jvlcek, line 66: I suggest adding a note that the trailing / is required.

`Note: The trailing / for the `/ui/sevice/` entry is required.`

PR #1476, opened 2020-08-10
Add Provider Tag Mapping to Capabilities Matrix

chessbyte:
Fixes https://github.com/ManageIQ/manageiq-documentation/issues/1436


capabilities_matrix/_topics/cloud_providers.md
```diff
@@ -54,6 +54,8 @@ The following table outlines the capabilities of {{ site.data.product.title_shor
 | Create/Update/Delete Network               | No                        | No              | No                          | Yes                        | No                                    |
 | Create/Update/Delete Subnet                | No                        | No              | No                          | Yes                        | No                                    |
 | Create/Update/Delete Routers               | Yes                       | Yes             | No                          | Yes                        | No                                    |
+| Tag Mapping from Provider                  | Yes                       | Yes             | No                          | Yes                        | Yes?                                  |
```
gtanzillo, line 57: This looks correct. I'm not sure about "IBM PowerVC", either. But I'm assuming it's also supported if it's using the same Openstack provider backend. Perhaps @agrare can confirm.

chessbyte: Changed to `Yes` per @agrare 
We can always fix later if we find out it is not working.

PR #1484, opened 2020-10-01
Remove files that are not linked to

bdunne:
:scissors: :fire: :toilet: 

The index.md in these directories doesn't link to these files.  Also I would expect people to use the updated version in https://github.com/ManageIQ/manageiq-documentation/blob/master/_includes/installing-vmware-vddk.md instead of the outdated content here.


Fryguy (approved):

Fryguy:
LGTM, @dehawkins512 can you also review?

Issue #1449, opened 2020-05-28
Migration Path from Appliance to Podified


chessbyte:
- [x] Document in manageiq-documentation a manual process for a basic transfer/upgrade (https://github.com/ManageIQ/manageiq-documentation/issues/1449#issuecomment-689675253) https://github.com/ManageIQ/manageiq-documentation/pull/1486

Continued in https://github.com/ManageIQ/manageiq-documentation/issues/1510

bdunne:
Initial steps for migration path:
```
On the appliance:
DB backup
  pg_dump -Fc -d vmdb_production > /root/pg_dump
get key from certs/v2_key and Base64 it for Openshift
  rails r "puts Base64.encode64(ManageIQ::Password.v2_key.to_s)"
get the region number (this will need to be set in the CR if it's not 0)
  rails r "puts MiqRegion.my_region.region"


Do the prereqs (create: CRD, role, role binding, service account, operator)
edit the CR
  - set the application domain (required)
  - set the region number if it's not 0 (the default)
oc create -f the CR
oc edit secret app-secrets
  replace "encryption-key" with the value from the appliance above
oc debug pod/orchestrator-6f65876dc5-2tsx8
  Keep this running...
oc edit deployment/orchestrator adding...
  spec:
    template:
      spec:
        nodeSelector:
          kubernetes.io/hostname: nope (anything invalid will work, we want to prevent it from starting)
oc delete replicaset.apps/orchestrator-6f65876dc5 (the old one, the new pod will be "pending")
back in the debug pod:
  cd /var/www/miq/vmdb
  source ./container_env
  DISABLE_DATABASE_ENVIRONMENT_CHECK=1 rake db:drop db:create
oc rsh to the database pod
  cd /var/lib/pgsql
  download your backup here
  pg_restore -d vmdb_production pg_dump
  exit
back in the debug pod:
  rails r "MiqServer.first.update!(:guid => ENV['GUID'])"
  rake db:migrate (just in case?)
  exit
oc edit deployment/orchestrator removing...
  spec:
    template:
      spec:
        nodeSelector:
oc delete replicaset.apps/orchestrator-8bd47c86 (the one with the pending pod)
```

chessbyte:
@bdunne It would be good to formalize this to a new document in https://www.manageiq.org/docs/reference/ perhaps in the `Installation` section or (if you are bold) in a new section called `Upgrades`

blancett:
[...]

PR #1486, opened 2020-10-05
Add documentation for migrating from appliance(s) to pods

bdunne:
Phase 1 of https://github.com/ManageIQ/manageiq-documentation/issues/1449

### DIFF
--- a/installing_on_kubernetes/_topics/migrate_from_appliance.md
+++ b/installing_on_kubernetes/_topics/migrate_from_appliance.md
@@ -1,0 +1,89 @@
+## Migrating from Appliances
+
+### Collect data from the appliance
+1. Take a backup of the database
+
+        $ pg_dump -Fc -d vmdb_production > /root/pg_dump
+
+2. Export the encryption key and Base64 encode it for the Kubernetes Secret.
+
+        $ vmdb && rails r "puts Base64.encode64(ManageIQ::Password.v2_key.to_s)"
+
+3. Get the region number
+
+        $ vmdb && rails r "puts MiqRegion.my_region.region"
+
+4. Get the GUID of the server that you want to run as.
+
+        $ vmdb && cat GUID
+
+### Restore the backup into the kubernetes environment
+1. Create a YAML file defining the Custom Resource (CR). Minimally you'll need the following:
+
+        apiVersion: {{ site.data.product.operator_namespace }}/{{ site.data.product.operator_api_version }}
+        kind: {{ site.data.product.operator_custom_resource_kind }}
+        metadata:
+          name: <friendly name for you CR instance>
+        spec:
+          applicationDomain: <your application domain name>
+          databaseRegion: <region number from the appliance above>
+          serverGuid: <GUID value from appliance above>
+
+2. Create the CR in your namespace. Once created, the operator will create several additional resources and start deploying the app.
+
+        $ oc create -f <file name from above>
+
+3. Edit the app secret inserting the encryption key from the appliance. Replace the "encryption-key" value with the value we exported from the appliance above.
+
+        $ oc edit secret app-secrets
+
+4. Find the orchestrator pod and start a debug session into it. Keep this running in the background...
+
+        $ oc get pods -o name | grep orchestrator
+        $ oc debug pod/orchestrator-123456abcd-789ef
+
+5. Temporarily prevent the orchestrator from starting by adding the following to the deployment:
+
+        $ oc edit deployment/orchestrator
+
+        spec:
+          template:
+            spec:
+              nodeSelector:
+                kubernetes.io/hostname: nope
+
+6. Delete the old replica set, the new one will sit in "pending" state.
+
+        $ oc delete replicaset.apps/orchestrator-123456abcd
+
+7. Back in the debug pod from step 4:
+
+        $ cd /var/www/miq/vmdb
+        $ source ./container_env
+        $ DISABLE_DATABASE_ENVIRONMENT_CHECK=1 rake db:drop db:create
+
+8. `oc rsh` into the database pod and restore the database backup
+
+        $ cd /var/lib/pgsql
+        # --- download your backup here ---
+        $ pg_restore -d vmdb_production <your_backup_file>
+        $ rm -f <your_backup_file>
+        $ exit
+
+9. Back in the debug pod from step 4:
+
+        $ rake db:migrate
+        $ exit
+
+10. Delete the node selector that we added above `oc edit deployment/orchestrator` removing:
+
+        spec:
+          template:
+            spec:
+              nodeSelector:
+
+11. Delete the pending orchestrator deployment
+
+        $ oc delete replicaset.apps/orchestrator-98765cba
+
+Done! The orchestrator will start deploying the rest of the pods required to run the application.
--- a/_data/product.yml
+++ b/_data/product.yml
@@ -5,3 +5,8 @@ title_short_l: manageiq
 title_abbr: miq
 title_abbr_uc: MIQ
 appliance_rpm_dir: /opt/manageiq/manageiq-appliance
+operator_api_version: v1alpha1
+operator_custom_resource_definition_name_plural: manageiqs
+operator_custom_resource_kind: ManageIQ
+operator_directory: https://github.com/ManageIQ/manageiq-pods/tree/master/manageiq-operator
+operator_namespace: manageiq.org
--- a/installing_on_kubernetes/_topics/prepare_kubernetes_namespace.md
+++ b/installing_on_kubernetes/_topics/prepare_kubernetes_namespace.md
@@ -1,0 +1,17 @@
+## Preparing the Kubernetes namespace
+The deploy directory referenced below can be found [here]({{ site.data.product.operator_directory }}/deploy)
+
+1. Search for the Custom Resource Definition (CRD) and create it if it doesn't already exist.
+
+        $ oc get crds | grep {{ site.data.product.operator_custom_resource_definition_name_plural }}.{{ site.data.product.operator_namespace }}
+        $ oc create -f deploy/crds/{{ site.data.product.operator_namespace }}_{{ site.data.product.operator_custom_resource_definition_name_plural }}_crd.yaml
+
+2. Set up RBAC.
+
+        $ oc create -f deploy/role.yaml
+        $ oc create -f deploy/role_binding.yaml
+        $ oc create -f deploy/service_account.yaml
+
+3. Deploy the operator in your namespace.
+
+        $ oc create -f deploy/operator.yaml
--- a/installing_on_kubernetes/index.md
+++ b/installing_on_kubernetes/index.md
@@ -1,0 +1,8 @@
+---
+---
+
+# Installing on Kubernetes / Openshift
+
+{% include_relative _topics/prepare_kubernetes_namespace.md %}
+
+{% include_relative _topics/migrate_from_appliance.md %}


installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,94 @@
+## Preparing for deployment
+
+1. Deploy the CRD (if it doesn't already exist)
+`$ oc get crds | grep manageiqs.manageiq.org`
```
jvlcek, line 4: I suggest having the example commands placed under the text by using the **```bash**  header.

You also might want to consider listing the command used to determine if the CRD already exists separately from the one used to create it if it does not.

see: https://github.com/ManageIQ/manageiq-pods/tree/master/manageiq-operator#running-manageiq-under-operator-control

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,94 @@
...
+`$ oc create -f deploy/crds/manageiq.org_manageiqs_crd.yaml`
+
+2. Set up RBAC
+```bash
+$ oc create -f deploy/role.yaml
+$ oc create -f deploy/role_binding.yaml
+$ oc create -f deploy/service_account.yaml
+```
+
+3. Deploy the operator
```
jvlcek, line 14: The below example depicts how to run the latest ManageIQ Operator image. I suggest considering the following wording:
**_Run the latest ManageIQ Operator image from the registry in the cluster_**

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,94 @@
...
+`$ oc create -f deploy/operator.yaml`
```
jvlcek, line 15: Again I suggest having the example commands placed under the text by using the **```bash**    header.

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Migrating from Appliances
+
+### Collect data from the appliance
```
jvlcek, line 19: Again I suggest having the example commands placed under the text by using the ```bash header.

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,94 @@
...
+`vmdb && rails r "puts Base64.encode64(ManageIQ::Password.v2_key.to_s)"`
+
+3. Get the region number
+`vmdb && rails r "puts MiqRegion.my_region.region"`
+
+4. Get the GUID of the server that you want to run as.
+`vmdb && cat GUID`
+
+### Restore the backup into the kubernetes environment
+1. Create a YAML file defining the CR. Minimally you'll need the following:
```
jvlcek, line 33: Please introduce acronyms by using the full name, at least for the first reference of CR, CDR... 
e.g.: _**Custom Resource (CR)**_

jvlcek, line 33: Perhaps recommending a file name here so it can be referenced below.

Create a YAML file, **my_custom_resource.yaml**.,  defining the CR. Minimally you'll need the following:

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,94 @@
...
+metadata:
+  name: miq
+spec:
+  applicationDomain: <your application domain name>
+  databaseRegion: <region number from the appliance above>
+  serverGuid: <GUID value from appliance above>
+```
+
+2. Create the CR in your namespace. Once created, the operator will create several additional resource and start deploying the app.
+`oc create -f <file name from above>`
```
jvlcek, line 46: Suggested example:

` ```bash`
`# oc create -f <file name from above> `
`oc create -f my_custom_resource.yaml `
` ``` `

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,94 @@
...
+
+3. Edit the app secret inserting the encryption key from the appliance. Replace the "encryption-key" value with the value we exported from the appliance above.
+`oc edit secret app-secrets`
```
jvlcek, line 49: Perhaps a more complete example, which might also include, showing how the secret is created.

e.g.:
` ```bash `
`  oc create secret app-secrets  \ `
`  --from-literal=encryption-key=<value exported from the application above> `  
` ``` `

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,94 @@
...
+
+4. Start a debug session into the orchestrator. Keep this running in the background...
+`oc debug pod/orchestrator-6f65876dc5-2tsx8`
```
jvlcek, line 52: Again I suggest having the example commands placed under the text by using the **```bash** header.

And perhaps show how to get the orchestrator pod
e.g.:

` ```bash `
`  $ oc login -u admin -p <pw> https://<my-container-host>/ `
`  $ oc get pods -n my-miq-project | grep orchestrator `
`     orchestrator-abcdef123-12345         1/1     Running   0          22d `
` $ `
` $ #  oc debug pod/<orchestrator pod name from above output> `
` $ oc debug pod/orchestrator-abcdef123-12345`
` ``` `

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,94 @@
...
+
+5. Temporarily prevent the orchestrator from starting by adding the following to the deployment:
+`oc edit deployment/orchestrator`
```
jvlcek, line 55: Suggest using the **```bash** header

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,94 @@
...
+```yaml
+spec:
+  template:
+    spec:
+      nodeSelector:
+        kubernetes.io/hostname: nope (anything invalid will work, we want to prevent it from starting)
+```
+
+6. Delete the old replica set, the new one will sit in "pending" state.
+`oc delete replicaset.apps/orchestrator-6f65876dc5`
```
jvlcek, line 65: Suggest using the ```bash header

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,94 @@
...
+
+7. Back in the debug pod:
```
jvlcek, line 67: Could step 4 `Start a debug session...` be done here, combining step 4 into a single step here?

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,94 @@
...
+```bash
+cd /var/www/miq/vmdb
+source ./container_env
+DISABLE_DATABASE_ENVIRONMENT_CHECK=1 rake db:drop db:create
+exit
+```
+
+8. `oc rsh` into the database pod and restore the database backup
+```bash
```
jvlcek, line 76: Then this example can include how to oc rsh into the database pod.

I suggest considering something like this:

e.g.:
` $ oc login -u admin -p <pw> https://<my-container-host>/ `
`$ oc get pods -n my-miq-project | grep postresql `
`postgresql-abcdef123-12345 1/1 Running 0 22d `
` $ `
` $ # oc debug pod/<database pod name from above output> `
` $ oc rsh postgresql-abcdef123-12345 `

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,94 @@
...
+cd /var/lib/pgsql
```
jvlcek, line 77: Suggest using the **```bash** header

jvlcek (changes requested):
Hey @bdunne Good job getting this written down.

I've posted some suggestions for you to consider.

JoeV

installing_on_kubernetes/_topics/migrate_from_appliance.md
bdunne: Nope.  The debug pod needs to be started before we remove the deployment, but other steps can't be done until that deployment is removed.  Otherwise they'll conflict

installing_on_kubernetes/_topics/migrate_from_appliance.md
bdunne: :+1:

bdunne:
Okay, I ran through the steps again and worked out some formatting issues.  @jvlcek Want to re-review?

Fryguy:
> Phase 1 of #1449

1 out of 1449?!...you got a long way to go 🤣 🤣 🤣

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,131 @@
+## Preparing the Kubernetes namespace
+The deploy directory referenced below can be found in [the manageiq-pods repo](https://github.com/ManageIQ/manageiq-pods/tree/master/manageiq-operator/deploy)
```
Fryguy, line 2: This document is going to be a pain for productization... wondering how we can do that.  At a minimum perhaps we can move things like manageiq specific pod namings into https://github.com/ManageIQ/manageiq-documentation/blob/master/_data/product.yml (or use what's there?)

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,131 @@
...
+
+1. Search for the Custom Resource Definition (CRD) and create it if it doesn't already exist.
+```bash
```
Fryguy, line 5: if you indent these blocks 3 spaces, they will render as part of number 1 instead of after it.  Not sure if we follow this pattern elsewhere in the docs.

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Migrating from Appliances
+
+### Collect data from the appliance
+1. Take a backup of the database
+```bash
+pg_dump -Fc -d vmdb_production > /root/pg_dump
+```
+
+2. Export the encryption key and Base64 encode it for Openshift
```
Fryguy, line 32: OpenShift should have a capital S.  Additionally, we probably should be less OpenShift oriented where possible, so can this say Kubernetes instead (or Kubernetes/OpenShift)

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,131 @@
...
+
+3. Get the region number
+```bash
+vmdb && rails r "puts MiqRegion.my_region.region"
+```
+
+4. Get the GUID of the server that you want to run as.
+```bash
+vmdb && cat GUID
+```
```
Fryguy, line 45: Perhaps not for this doc, but I wonder if we can have a rake task that just dumps these values..something like:

```bash
$ vmdb && rake evm:export:identity
dVhmSWdTQVVxNU96OGdvYy96SThIT09vMFNJKytTZDltZnBnQmFuWUlNND0=
0
252ca2e0-ebc9-0138-6e61-7affc243f4d3
```

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,131 @@
...
+  databaseRegion: <region number from the appliance above>
+  serverGuid: <GUID value from appliance above>
+```
+
+2. Create the CR in your namespace. Once created, the operator will create several additional resource and start deploying the app.
+```bash
+oc create -f <file name from above>
+```
+
+3. Edit the app secret inserting the encryption key from the appliance. Replace the "encryption-key" value with the value we exported from the appliance above.
```
Fryguy, line 65: Can this be done as step one?  That was the app-secrets already exists?

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+7. Back in the debug pod from step 4:
+```bash
+cd /var/www/miq/vmdb
+source ./container_env
+DISABLE_DATABASE_ENVIRONMENT_CHECK=1 rake db:drop db:create
+```
+
+8. `oc rsh` into the database pod and restore the database backup
```
Fryguy, line 103: We might want an analogous kubectl command otherwise this requires installing oc

Fryguy:
This looks really good @bdunne ... As we move forward, I'm thinking from a deployment perspective, it's really weird to have to essentially "pause" the orchestrator right after you deploy it.  I'm thinking this might be better done as a value in the CR.  So when you deploy the CR, you would set something like `databaseRestoreMode: true` which would let the operator deploy everything but then put the orchestrator into a paused state or, even better, not launch it at all and we have a user deploy a "restore" pod/job instead.  Then to "unpause" the operator you would remove that value from the CR, and the operator would reconcile that change to deploy the orchestrator.

bdunne:
> This looks really good @bdunne ... As we move forward, I'm thinking from a deployment perspective, it's really weird to have to essentially "pause" the orchestrator right after you deploy it. I'm thinking this might be better done as a value in the CR. So when you deploy the CR, you would set something like `databaseRestoreMode: true` which would let the operator deploy everything but then put the orchestrator into a paused state or, even better, not launch it at all and we have a user deploy a "restore" pod/job instead. Then to "unpause" the operator you would remove that value from the CR, and the operator would reconcile that change to deploy the orchestrator.

Yeah, I was thinking about something similar.  I was thinking create a secret with:
```yaml
applianceHost: appliance.example.com
applianceUser: root
appliancePassword: smartvm
```
Then put `deploymentState: applianceImport` in the CR and the operator can monitor/update that field and fire off an import job that would fetch everything.  Once complete, switch to  `deploymentState: run` and the Orchestrator would be deployed and app would start up.

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -0,0 +1,131 @@
...
+kind: ManageIQ
+metadata:
+  name: miq
+spec:
+  applicationDomain: <your application domain name>
+  databaseRegion: <region number from the appliance above>
+  serverGuid: <GUID value from appliance above>
+```
+
+2. Create the CR in your namespace. Once created, the operator will create several additional resource and start deploying the app.
```
gtanzillo, line 60: "resource" should be "resources"

installing_on_kubernetes/_topics/migrate_from_appliance.md
bdunne: The operator creates the secret after Step 2.  Alternatively we could manually define the secret, but I think this might be easier until this is fully automated.

installing_on_kubernetes/_topics/migrate_from_appliance.md
bdunne: Most of these can be `s/oc/kubectl/`, but there are some that aren't that simple (oc debug).  I'd rather stick with one CLI tool until then.

installing_on_kubernetes/_topics/migrate_from_appliance.md
Fryguy: yeah i was thinking manually define the secret (or offer a yaml that people could edit, then apply)

Fryguy:
Only comment I have left is on the productization aspects.  If possible, let's remove any manageiq references, and where we can't use product.yml and `{{ }}` substitution instead.

bdunne:
> Only comment I have left is on the productization aspects. If possible, let's remove any manageiq references, and where we can't use product.yml and `{{ }}` substitution instead.

:+1: Done

Fryguy (approved):
LGTM.. @jvlcek you have this marked as "Request changes"...are you good with it now?

jvlcek (approved):

jvlcek:
LGTM 👍

PR #1491, opened 2020-10-11
Add IBM Cloud Power Virtual Servers Provider Documentation

jaywcarman:
Initial documentation for the IBM Cloud Power Virtual Servers provider.

The [manageiq-providers-ibm_cloud](https://github.com/ManageIQ/manageiq-providers-ibm_cloud) plugin initially supports two IBM Cloud services: Power Virtual Servers and VPC. This PR covers only Power Virtual Servers.


jaywcarman:
FYI: @dehawkins512 @jwcroppe

capabilities_matrix/_topics/cloud_providers.md
```diff
@@ -2,60 +2,60 @@
...
+| ------------------------------------------ | ------------------------- | --------------- | --------------------------- | -------------------------- | ------------------------------------- | --------------------------------- |
+| Relationship Discovery                     | Yes                       | Yes             | Yes                         | Yes                        | Yes                                   | Yes                               |
+| Instance Drift Comparison                  | Yes                       | Yes             | No                          | Yes                        | No                                    | No                                |
+| Track Instance Genealogy                   | Yes                       | Yes             | Yes                         | Yes                        | Yes                                   | No                                |
+| Capacity & Utilization                     | Yes                       | Yes             | Yes                         | Yes                        | No                                    | No                                |
```
jwcroppe, line 10: Are we sure this is no?  We don't have utilization data, but we have consumed data I believe for CPU, memory and disk?

capabilities_matrix/_topics/cloud_providers.md
```diff
@@ -2,60 +2,60 @@
...
+| Discovery - Provider                       | No                        | No              | No                          | No                         | No                                    | No                                |
+| Disk Addition to Instance                  | Yes                       | Yes             | No                          | Yes                        | No                                    | Yes                               |
+| Key Pairs Inventory                        | Yes                       | No              | Yes                         | Yes                        | Yes                                   | Yes                               |
+| Key Pairs Management                       | Yes                       | No              | No                          | Yes                        | Yes                                   | Yes                               |
+| Reporting                                  | Yes                       | Yes             | Yes                         | Yes                        | No                                    | No                                |
```
jwcroppe, line 17: Are we sure reporting wouldn't "just work"?

capabilities_matrix/_topics/cloud_providers.md
```diff
@@ -2,60 +2,60 @@
...
+| Right Sizing                               | Yes                       | Yes             | Yes                         | Yes                        | Yes                                   | No                                |
+| Chargeback                                 | Yes                       | Yes             | Yes                         | Yes                        | No                                    | Yes                               |
+| Remote Console Instance Access             | No                        | No              | No                          | Yes                        | Yes (for NovaLink-managed hosts only) | No                                |
+| Snapshot Creation and Removal              | No                        | No              | No                          | Yes                        | Yes                                   | No                                |
+| Instance Compliance Enforcement            | Yes                       | Yes             | Yes                         | Yes                        | No                                    | No                                |
```
jwcroppe, line 22: Would this "just work"?

capabilities_matrix/_topics/cloud_providers.md
```diff
@@ -2,60 +2,60 @@
...
+| Integrate with Service Catalogs            | Yes                       | Yes             | Yes                         | Yes                        | Yes                                   | Yes                               |
+| Instance Reconfiguration                   | No                        | No              | No                          | Yes                        | Yes                                   | No                                |
+| Volume Inventory                           | No                        | No              | Yes                         | Yes                        | Yes                                   | Yes                               |
+| Volume Creation/Deletion                   | No                        | No              | No                          | Yes                        | Yes                                   | Yes                               |
+| Instance Migration                         | No                        | No              | No                          | Yes                        | Yes                                   | No                                |
```
jwcroppe, line 33: Should we say N/A as this isn't supported by PowerVS - no somewhat gives the implication that it's a gap?

capabilities_matrix/_topics/cloud_providers.md
```diff
@@ -2,60 +2,60 @@
...
+| Provision from Volume to Instance          | No                        | No              | No                          | Yes                        | No                                    | No                                |
+| Provision from Volume Snapshot to Instance | No                        | No              | No                          | Yes                        | No                                    | No                                |
+| Orchestration Template Provisioning        | Yes                       | Yes             | No                          | Yes                        | No                                    | No                                |
+| Cloud-Init                                 | No                        | No              | No                          | Yes                        | No                                    | Yes                               |
+| Sysprep Windows Templates                  | No                        | No              | No                          | Yes                        | No                                    | No                                |
```
jwcroppe, line 44: Should we say N/A as this isn't supported by PowerVS - no somewhat gives the implication that it's a gap?

capabilities_matrix/_topics/cloud_providers.md
```diff
@@ -2,60 +2,60 @@
...
+| Relationship Discovery                     | Yes                       | Yes             | Yes                         | Yes                        | No                                    | Yes                               |
+| Cloud Network Inventory                    | Yes                       | Yes             | Yes                         | Yes                        | No                                    | Yes                               |
+| Cloud Subnet Inventory                     | Yes                       | Yes             | Yes                         | Yes                        | No                                    | Yes                               |
+| Network Router Inventory                   | No                        | No              | Yes                         | Yes                        | No                                    | No                                |
+| Security Groups                            | Yes                       | Yes             | Yes                         | Yes                        | No                                    | No                                |
```
jwcroppe, line 50: Should we say N/A as this isn't supported by PowerVS - no somewhat gives the implication that it's a gap?

capabilities_matrix/_topics/cloud_providers.md
```diff
@@ -2,60 +2,60 @@
...
+| Floating IP Addresses                      | Yes                       | Yes             | Yes                         | Yes                        | No                                    | No                                |
```
jwcroppe, line 51: Should we say N/A as this isn't supported by PowerVS - no somewhat gives the implication that it's a gap?

jwcroppe (commented):
Looks great!  A few minor questions.

capabilities_matrix/_topics/cloud_providers.md
jwcroppe: (Migration)

capabilities_matrix/_topics/cloud_providers.md
jwcroppe: (Compliance enforcement)

capabilities_matrix/_topics/cloud_providers.md
jwcroppe: (Sysprep Windows)

capabilities_matrix/_topics/cloud_providers.md
jwcroppe: (Security Groups)

capabilities_matrix/_topics/cloud_providers.md
jwcroppe: (Floating IPs)

capabilities_matrix/_topics/cloud_providers.md
jaywcarman: I think there's a distinction between data collected through the inventory refresher and C&U collector. The C&U data is stored in special 'metrics' tables in the DB.

So I think your questions is: Can some of the C&U reporting be run using data from the 'vms' table? I don't know.

capabilities_matrix/_topics/cloud_providers.md
jaywcarman: Since the 'IBM PowerVC via the OpenStack API' column is marked `No` I'm assuming this includes reporting C&U information.

capabilities_matrix/_topics/cloud_providers.md
jaywcarman: Here's the documentation I read to understand this capability: https://www.manageiq.org/docs/reference/latest/policies_and_profiles_guide/index.html

Things like enforcing an sshd configuration inside the VM would require direct access to the VM OS (maybe there's a way to add a MIQ ssh key or something?).

I was also looking at the 'IBM PowerVC via the OpenStack API' column, which is marked `No`, and assuming there's more to this capability than what we've implemented.

Guidance from the MIQ team would be helpful here.

capabilities_matrix/_topics/cloud_providers.md
jaywcarman: done

capabilities_matrix/_topics/cloud_providers.md
jaywcarman: done

capabilities_matrix/_topics/cloud_providers.md
jaywcarman: done

capabilities_matrix/_topics/cloud_providers.md
jaywcarman: done

capabilities_matrix/_topics/cloud_providers.md
jwcroppe: Sounds good, agree with "No" for now.

capabilities_matrix/_topics/cloud_providers.md
```diff
@@ -2,60 +2,60 @@
...
+| Key Pairs Inventory                        | Yes                       | No              | Yes                         | Yes                        | Yes                                   | Yes                               |
+| Key Pairs Management                       | Yes                       | No              | No                          | Yes                        | Yes                                   | Yes                               |
+| Reporting                                  | Yes                       | Yes             | Yes                         | Yes                        | No                                    | No                                |
+| Right Sizing                               | Yes                       | Yes             | Yes                         | Yes                        | Yes                                   | No                                |
+| Chargeback                                 | Yes                       | Yes             | Yes                         | Yes                        | No                                    | Yes                               |
```
jaywcarman, line 19: @gtanzillo Do you know if this Chargeback row is describing C&U + Allocation chargeback? Since the Power Virtual Servers provider will be capable of allocation based chargeback reports I marked it `Yes`, but that I don't think that fits of the "IBM PowerVC via the OpenStack API" column being marked `No`.

capabilities_matrix/_topics/cloud_providers.md
agrare: That's correct @jaywcarman , Reporting/chargeback can use consumed resources collected from ems_refresh but that is not the same as c&u/metrics

capabilities_matrix/_topics/cloud_providers.md
agrare: @jaywcarman on the InfraManager cap matrix SCVMM (which also doesn't support C&U) says "Yes (but limited to inventory allocation)" [[ref]](https://github.com/ManageIQ/manageiq-documentation/blob/master/capabilities_matrix/_topics/infrastructure_providers.md) so maybe we could use that

capabilities_matrix/_topics/cloud_providers.md
```diff
@@ -2,60 +2,60 @@
...
+| Floating IP Addresses                      | Yes                       | Yes             | Yes                         | Yes                        | No                                    | N/A                               |
+| Network Ports                              | Yes                       | Yes             | Yes                         | Yes                        | No                                    | Yes                               |
+| Load Balancer Inventory                    | Yes                       | Yes             | Yes                         | No                         | No                                    | No                                |
+| Create/Update/Delete Network               | No                        | No              | No                          | Yes                        | No                                    | No                                |
+| Create/Update/Delete Subnet                | No                        | No              | No                          | Yes                        | No                                    | No                                |
```
agrare, line 55: We say support create and delete in https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/network_manager/cloud_subnet.rb#L2-L4
is the No because create looks like it isn't implemented (is that a bug?) or because update isn't supported?

agrare (commented):
This looks awesome @jaywcarman !

One question about the cloud subnet CRUD and we can say "Allocation only" on the Chargeback

capabilities_matrix/_topics/cloud_providers.md
jaywcarman: :+1: That looks good to me - done.

capabilities_matrix/_topics/cloud_providers.md
jaywcarman: Good catch @agrare. Create and delete to do work* (and update is not implemented). The create form is still pretty messy - lots of n/a fields, missing PowerVS parameters, and `Subnet CIDR` is required to submit but not used by the API call for our "public" networks. We're thinking it would be better to disable subnet create for Kasparov than to try and document using the form in a confusing way. This [form is moving to DDF anyways](https://github.com/ManageIQ/manageiq-ui-classic/issues/6861).

I'll submit a PR to ManageIQ/manageiq-providers-ibm_cloud shortly to disable it.

agrare (approved):
:+1: LGTM

PR #1507, opened 2020-10-20
Updates for more URLS to relative links in ManageIQ

dehawkins512:


integration_with_openshift_container_platform/_topics/overview.md
```diff
@@ -1,10 +1,7 @@
 This guide walks you through adding a Red Hat OpenShift Container Platform cluster to a {{ site.data.product.title }} container provider catalog. This deployment focuses on enabling the OpenShift Container Platform cluster metrics plug-in, so that {{ site.data.product.title_short }} can collect information from an OpenShift Container Platform cluster upon integration.
 
-Each procedure in this guide is covered in greater detail in the
-[{{ site.data.product.title }}](https://www.ibm.com/support/knowledgecenter/SSFC4F_2.0.0/mcm/infrastructure/infra_mgmt_intro.html)
+Each procedure in this guide is covered in greater detail in the [{{ site.data.product.title }}](https://www.ibm.com/support/knowledgecenter/SSFC4F_2.0.0/mcm/infrastructure/infra_mgmt_intro.html)
```
chessbyte, line 3: if this doc is not in upstream, would it make sense to move it over at some point?

chessbyte:
why is `managing_providers/_topics/automation_management_providers.md` being deleted in this PR?

dehawkins512:
@chessbyte This has to be a mistake I will correct.  I updated the file and replaced many instances of <div class="note"> with markdown -> **Note:**

integration_with_openshift_container_platform/_topics/overview.md
dehawkins512: @chessbyte Yes. or reference another topic in upstream and I can make other changes downstream. It should be easy to create this doc upstream.

Fryguy: For the purposes of this particular PR I"m ok with this as it was the same before, but I agree we need to change this part to avoid referencing the IBM docs if possible.

integration_with_openshift_container_platform/_topics/prerequisites.md
```diff
@@ -1,10 +1,7 @@
 This guide assumes that you have:
 
-  - [Already deployed {{ site.data.product.title }} on your chosen
-    platform](https://www.ibm.com/support/knowledgecenter/SSFC4F_2.0.0/mcm/infrastructure/infra_mgmt_intro.html)
+  - [Already deployed {{ site.data.product.title }} on your chosen platform](https://www.ibm.com/support/knowledgecenter/SSFC4F_2.0.0/mcm/infrastructure/infra_mgmt_intro.html)
```
Fryguy, line 3: Same here.

Fryguy:
@chessbyte I'm not seeing the file deleted, it's just that GitHub collapsed the diff because that particular file was a large diff.

chessbyte:
ah ok - my bad

PR #1517, opened 2020-12-03
Add notes about multi-server multi-zone appliance -> kubernetes migration

bdunne:
Part of https://github.com/ManageIQ/manageiq-documentation/issues/1510

### DIFF
--- a/installing_on_kubernetes/_topics/migrate_from_appliance.md
+++ b/installing_on_kubernetes/_topics/migrate_from_appliance.md
@@ -1,5 +1,11 @@
 ## Migrating from Appliances
 
+### Notes
+- Current architecture in podified limits us to running a single server and zone in a Kubernetes namespace.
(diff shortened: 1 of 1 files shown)


installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -1,5 +1,11 @@
 ## Migrating from Appliances
 
+### Notes
+- Current architecture in podified limits us to running a single server and zone in a Kubernetes namespace.
```
chessbyte, line 4: I thought that we would leverage Kubernetes node labels to allow for multiple zones.  Or what that work never completed?  /cc @Fryguy

bdunne: I believe that is the design idea, but I don't think it is implemented yet.

Issue #1510, opened 2020-11-03
Migration Path from Appliance to Podified - continued for Lasker release


gtanzillo:
Continued from https://github.com/ManageIQ/manageiq-documentation/issues/1449

- [x] Document how to deal with multi-server/multi-zone deployments. https://github.com/ManageIQ/manageiq-documentation/pull/1517
- [x] Document how to deal with multi-region deployments. #1521

PR #1521, opened 2020-12-07
Add notes about multi-region appliance -> kubernetes migration

bdunne:
Part of #1510


installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -1,11 +1,25 @@
 ## Migrating from Appliances
 
 ### Notes
-- Current architecture in podified limits us to running a single server and zone in a Kubernetes namespace.
+
+- Multi-server / multi-zone:
+  Current architecture in podified limits us to running a single server and zone in a Kubernetes namespace.
```
Fryguy, line 6: I think technically we do have multi-zone - it can be configured via node labels.  I don't remember the specifics, so we'd have to dig in.

Fryguy:
Going to merge anyway, and we can update this with more accurate information about multi-zone in a follow up.

PR #1535, opened 2021-03-03
[general_configuration] Remove auth docs for Zones

NickLaMuro:
This is to be paired with ManageIQ/manageiq-ui-classic#7650 which will remove this feature from the UI.


Fryguy (approved):

Fryguy:
ManageIQ/manageiq-ui-classic#7650  is merged

PR #1542, opened 2021-03-04
Minor Fixes to Capabilities

chessbyte:
Also removed a few rows that were all red.


chessbyte:
@agrare Did we also remove infrastructure provider discovery?  I recall discussing that recently, but do not remember where we ended up with that.

agrare (approved):

PR #1560, opened 2021-04-21
Remove topics related to "Discovering Providers" from docs

dehawkins512:
Product capability to Discover Infrastructure Providers was removed..  cleaning up the docs.


agrare:
@dehawkins512 I also see `_includes/openstack/discovering-openstack-infrastructure-providers.md` which talks about `scans a network segment and searches for the Bare Metal (Ironic) service` which we should also remove

dehawkins512:
> 
> 
> @dehawkins512 I also see `_includes/openstack/discovering-openstack-infrastructure-providers.md` which talks about `scans a network segment and searches for the Bare Metal (Ironic) service` which we should also remove

@agrare Thanks.  I also see these topics:
Discovering Amazon EC2 Cloud Providers
Discovering Azure Providers
We can delete these as well, yes?

PR #1567, opened 2021-04-27
Update the OCP Add Container Docs for Alerts/Metrics

agrare:
The documentation needs to be updated after migrating the new container provider screen to DDF, specifically the "Detect" route button is no longer available so we should document what `oc` command to run in order to get the hostname.

Also we should indicate that Prometheus Alerts are not supported on OCPv4.

This is what the current dialogs look like:

Default Endpoint:
Screenshot from 2021-04-27 12-49-50

Metrics:
Screenshot from 2021-04-27 12-49-44

Alerts:
Screenshot from 2021-04-27 12-49-47

### DIFF
--- a/_includes/provider-ocp-add-container.md
+++ b/_includes/provider-ocp-add-container.md
@@ -10,18 +10,7 @@
 5.  Enter the appropriate **Zone** for the provider. If you do not
     specify a zone, it is set to `default`.
 
-6.  From the **Alerts** list, select **Prometheus** to enable external
-    alerts. Selecting **Prometheus** adds an **Alerts** tab to the lower
-    pane to configure the Prometheus service. Alerts are disabled by
-    default.
-
-7.  From the **Metrics** list, select **Hawkular** or **Prometheus** to
-    collect capacity and utilization data, or leave as **Disabled**.
-    Selecting **Prometheus** or **Hawkular** adds a **Metrics** tab to
-    the lower pane for further configuration. Metrics are disabled by
-    default.
-
-8.  In the **Default** tab, configure the following for the OpenShift
+6.  In the **Default** tab, configure the following for the OpenShift
     provider:
 
     1.  Select a **Security Protocol** method to specify how to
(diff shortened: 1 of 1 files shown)


_includes/provider-ocp-add-container.md
```diff
@@ -10,18 +10,7 @@
...
-6.  From the **Alerts** list, select **Prometheus** to enable external
-    alerts. Selecting **Prometheus** adds an **Alerts** tab to the lower
-    pane to configure the Prometheus service. Alerts are disabled by
-    default.
-
-7.  From the **Metrics** list, select **Hawkular** or **Prometheus** to
-    collect capacity and utilization data, or leave as **Disabled**.
-    Selecting **Prometheus** or **Hawkular** adds a **Metrics** tab to
-    the lower pane for further configuration. Metrics are disabled by
-    default.
```
agrare, line 22: These drop-down no longer exist, they have been moved to endpoint tabs with the default type being "Disabled"

_includes/provider-ocp-add-container.md
```diff
@@ -73,10 +62,11 @@
     5.  Click **Validate** to confirm that {{ site.data.product.title_short }} can connect
         to the OpenShift Container Platform provider.
 
-9.  For the **Prometheus** alerts service, add the Prometheus alerts
-    endpoint in the **Alerts** tab:
+7.  In the **Alerts** endpoint tab optionally configure the alerts service:
 
-    1.  Select a **Security Protocol** method to specify how to
+    1.  Select a service type, or leave disabled.  Prometheus alerts are only supported on OCP v3
```
agrare, line 67: OCPv4 doesn't have the Prometheus Alert Buffer service installed, I've opened an issue to improve the error message if someone tries to enable this on an OCPv4 (https://github.com/ManageIQ/manageiq-providers-openshift/issues/208) but we should at least document this.

dehawkins512 (approved):

PR #1572, opened 2021-05-19
Shrink the header titles on the capabilities matrix

Fryguy:
- Also fixes some of the header sizes which are a bit off

@agrare Please reivew.  Since I added the providers section at the front, the full names are there, and they aren't needed in the individual tables.  Personally, I think this is easier to read...WDYT, since you are usually the one who has to edit this?


kbrock (approved):

agrare (approved):

Fryguy:
Skipping backport to `morphy`, because it is already in the branch.

Issue #1562, opened 2021-04-23
Bad method name in 'methods available for automation' for Kasparov & Jansa


eleanorcuallado:
We recenly switched from Ivanchuk to Kasparov and I noticed during the testing of our code that the documentation does not take into account the following change in ManageIQ:

https://github.com/ManageIQ/manageiq/pull/19211

which changed the name of the `display` attribute to `visible`. While the `display` method is still available, `display=` no longer exists (at least in Kasparov). The name should be changed to `visible=`.

(We are still going through our testing, I will update the issue if I find other similar errors).

PR #1587, opened 2021-08-05
Remove OpenStack reference from PowerVC

jaywcarman:
There is now a standalone PowerVC provider:
https://github.com/ManageIQ/manageiq-providers-ibm_power_vc

The new PowerVC provider is a subclass of the OpenStack provider. Though
it does still use the OpenStack API this is now transparent to the user.


jaywcarman:
I'm working through #1562 to _finally_ close out all of the items, and this was included. I had thought that wording was more like "via the OpenStack provider", but now I see that what's there, "via the OpenStack API", is still accurate even with the new provider (https://github.com/ManageIQ/manageiq-providers-ibm_power_vc). I thinks it's better to remove it (hence this patch) but don't feel strongly about it.

PR #1596, opened 2021-08-26
Add instructions on adding an OKE provider

agrare:
Depends:
- [x] https://github.com/ManageIQ/manageiq-providers-oracle_cloud/pull/30

https://github.com/ManageIQ/manageiq-providers-oracle_cloud/issues/31


Fryguy:
Skipping backport to `morphy`, because it is already in the branch.

PR #1597, opened 2021-08-30
[WIP] Rename service evmserverd to manageiq

kbrock:
Part of https://github.com/ManageIQ/manageiq/issues/21397

We are keeping an alias so old commands still work, but updating documentation to reference new process name.


kbrock:
this got stalled on the installation side.
Thinking maybe we want to just set an alias and be done with it

kbrock:
I had wanted to just add the alias, but it became a big endeavor.
Thinking just doing the alias in the short term and then worrying about documentation later

PR #1613, opened 2021-11-04
We can simplify the podified database restore

bdunne:
- We don't necessarily need to be unscheduleable, we just need the app to not be running which can be done by changing the command

### DIFF
--- a/installing_on_kubernetes/_topics/migrate_from_appliance.md
+++ b/installing_on_kubernetes/_topics/migrate_from_appliance.md
@@ -57,53 +57,33 @@
 
         $ oc edit secret app-secrets
 
-4. Find the orchestrator pod and start a debug session into it. Keep this running in the background...
+4. Temporarily hijack the orchestrator by adding the following to the deployment:
 
-        $ oc get pods -o name | grep orchestrator
-        $ oc debug pod/orchestrator-123456abcd-789ef
+        $ kubectl patch deployment orchestrator -p '{"spec":{"template":{"spec":{"containers":[{"name":"orchestrator","command":["sleep", "1d"]}]}}}}'
 
-5. Temporarily prevent the orchestrator from starting by adding the following to the deployment:
+5. Shell into the orchestrator container:
 
-        $ oc edit deployment/orchestrator
-
-        spec:
-          template:
-            spec:
-              nodeSelector:
-                kubernetes.io/hostname: nope
-
-6. Delete the old replica set, the new one will sit in "pending" state.
-
-        $ oc delete replicaset.apps/orchestrator-123456abcd
-
-7. Back in the debug pod from step 4:
+        $ oc rsh deploy/orchestrator
 
         $ cd /var/www/miq/vmdb
         $ source ./container_env
         $ DISABLE_DATABASE_ENVIRONMENT_CHECK=1 rake db:drop db:create
 
-8. `oc rsh` into the database pod and restore the database backup
+6. `oc rsh` into the database pod and restore the database backup
 
         $ cd /var/lib/pgsql
         # --- download your backup here ---
-        $ pg_restore -d vmdb_production <your_backup_file>
+        $ pg_restore -v -d vmdb_production <your_backup_file>
         $ rm -f <your_backup_file>
         $ exit
 
-9. Back in the debug pod from step 4:
+7. Back in orchestrator pod from step 5:
 
         $ rake db:migrate
         $ exit
 
-10. Delete the node selector that we added above `oc edit deployment/orchestrator` removing:
+8. Delete the orchestrator deployment to remove the command override that we added above:
 
-        spec:
-          template:
-            spec:
-              nodeSelector:
-
-11. Delete the pending orchestrator deployment
-
-        $ oc delete replicaset.apps/orchestrator-98765cba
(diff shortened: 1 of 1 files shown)


installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -57,53 +57,48 @@
 
         $ oc edit secret app-secrets
 
-4. Find the orchestrator pod and start a debug session into it. Keep this running in the background...
-
-        $ oc get pods -o name | grep orchestrator
-        $ oc debug pod/orchestrator-123456abcd-789ef
-
-5. Temporarily prevent the orchestrator from starting by adding the following to the deployment:
+4. Temporarily hikack the orchestrator by adding the following to the deployment:
```
chessbyte, line 60:
```suggestion
4. Temporarily hijack the orchestrator by adding the following to the deployment:
```

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -57,53 +57,48 @@
...
-        $ pg_restore -d vmdb_production <your_backup_file>
+        $ pg_restore -d vmdb_production <your_backup_file> # Add -v to see progress
         $ rm -f <your_backup_file>
         $ exit
 
-9. Back in the debug pod from step 4:
+7. Back in orchestrator pod from step 5:
 
         $ rake db:migrate
+        $ rails r "MiqWorker.destroy_all"
```
Fryguy, line 92: @bdunne What is this step for?  Aside, I thought this happened automatically on startup anyway?  (cc @agrare didn't we just talk about this?)

agrare: On evm_server startup we kill all workers: https://github.com/ManageIQ/manageiq/blob/master/lib/workers/evm_server.rb#L22

Fryguy: Oh I just realized not on morphy though?

Fryguy: If we don't kill the workers nicely in morphy, then I'd rather do the following for this one-off case as opposed to the changes in #205

```
rails r "MiqWorker; MiqWorker.inheritance_column = "__type_disabled"; MiqWorker.destroy_all"
```

agrare: I believe we still did it was just moved out of the rake task and into `lib/workers/evm_server.rb`: https://github.com/ManageIQ/manageiq/blob/morphy/lib/tasks/evm_application.rb#L21

Fryguy: Oh yeah you're right, and the delete code exists too https://github.com/ManageIQ/manageiq/blob/morphy/app/models/miq_server/worker_management/monitor/kill.rb#L13-L27

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -57,53 +57,33 @@
...
         $ source ./container_env
         $ DISABLE_DATABASE_ENVIRONMENT_CHECK=1 rake db:drop db:create
 
-8. `oc rsh` into the database pod and restore the database backup
+6. `oc rsh` into the database pod and restore the database backup
 
         $ cd /var/lib/pgsql
         # --- download your backup here ---
-        $ pg_restore -d vmdb_production <your_backup_file>
+        $ pg_restore -d vmdb_production <your_backup_file> # Add -v to see progress
```
Fryguy, line 76: Should we just make -v the default ... I can't imagine why you wouldn't want it

installing_on_kubernetes/_topics/migrate_from_appliance.md
```diff
@@ -57,53 +57,33 @@
...
-        spec:
-          template:
-            spec:
-              nodeSelector:
-
-11. Delete the pending orchestrator deployment
+8. Delete the orchestrator deployment to remove the command override that we added above:
 
-        $ oc delete replicaset.apps/orchestrator-98765cba
+        $ kubectl delete deployment orchestrator
```
Fryguy, line 87: If we delete the deployment what re-creates it? The operator? If so, will the GUID be lost and is that a problem? I thinking about the normal backup/restore case where you backup your own database and restore it, where you might want to keep the GUID.

bdunne: The operator will re-create it.  The GUID will not be lost, the `serverGuid` property can be set manually in the CR or the UID of the CR is used if none was provided.

Fryguy: Oh, UID of the _CR_.  That's the key for me.  Thanks.

installing_on_kubernetes/_topics/migrate_from_appliance.md
bdunne: Done

installing_on_kubernetes/_topics/migrate_from_appliance.md
bdunne: Fixed

Fryguy (approved):
cc @dehawkins512

PR #1616, opened 2021-11-23
Split up managing provider docs

agrare:
So high level I created a `managing_providers/infrastructure_providers` folder where I moved the provider specific (e.g. rhev/scvmm/vsphere) docs to a new page rather than simply `include_relative` into the top-level `managing_providers/index.md`

Then I moved the `_topics` referenced from those under `managing_providers/infrastructure_providers/_topics`

TODO:
- [x] Check other relative links like `../images`
- [x] Ensure the headers for the new pages are accurate


managing_providers/index.md
```diff
@@ -8,7 +8,7 @@
 In {{ site.data.product.title_short }}, a *provider* is an external virtualization, cloud, or containers environment that manages multiple virtual machines
 or instances residing on multiple hosts. One example is Red Hat Virtualization, a platform that manages multiple hosts and virtual machines.
 
-In {{ site.data.product.title_short }}, a *manager* is an external management environment that manages more than one type of resource. One example of a manager is OpenStack, which manages infrastucture, cloud, network, and storage resources.
+In {{ site.data.product.title_short }}, a *manager* is an external management environment that manages more than one type of resource. One example of a manager is OpenStack, which manages infrastructure, cloud, network, and storage resources.
```
agrare, line 11: Found this typo when my editor was suggesting "infrastucture" as an autocomplete option

agrare:
If we're good with this approach I'll continue with the rest of the provider types cloud/container/etc...

Fryguy:
cc @dehawkins512 for the changes that are coming.

_data/site_menu.yml
```diff
@@ -70,8 +70,94 @@ administration:
...
 
     - title: Managing Providers
-      path: "/docs/reference/latest/managing_providers/index.html"
-      desc: "Managing your infrastructure, cloud, and containers providers and datastores"
+      children:
+        - title: Overview
+          path: "/docs/reference/latest/managing_providers/index.html"
+          desc: "Managing your infrastructure, cloud, and containers providers and datastores"
+
+        - title: Red Hat Virtualization Providers
```
agrare, line 78: @Fryguy I was thinking of having these under e.g. `Infrastructure Providers`  sections but that'd be an extra level

chessbyte: to be accurate, it should be under the `Virtualization Infrastructure Providers` section (as opposed to `Physical Infrastructure Providers`).  I know that we use `Infrastructure Providers` in code and docs, but I believe that is a shorthand for `Virtualization Infrastructure Providers`

agrare: We call it `Infrastructure` and `Physical Infrastructure` in the UI, while I agree `Virtual Infrastructure` is more accurate I think the docs should match the UI so people aren't confused looking for the `Virtual Infrastructure` section in the UI
Screenshot from 2021-12-03 09-01-21

chessbyte: wondering if we should rename UI to have `Virtual Infrustructure` next to `Physical Infrastructure`

agrare: :+1: yeah I'd be for that

chessbyte: @Fryguy what do you think?  And this suggestion is not necessarily for this PR

Fryguy: Yeah I like "Virtual Infrastructure" in the UI and in the docs

chessbyte:
@agrare @Fryguy I am assuming something in this PR broke the docs.  See https://github.com/ManageIQ/manageiq.github.io/runs/4461266536

PR #1617, opened 2021-11-24
Use OS certificate bundle

kbrock:
Part of https://github.com/ManageIQ/manageiq-appliance/pull/341

This gets database ssl keys out of `/root`, which is only accessible to `root` user, and into a standard location which is accessible to other users, namely user `manageiq`

1. Store the key that signed the postgres server in the standard location
2. No longer store `root.crt` in `/root` so user manageiq can use it
3. Still configure postgres for public and private key
4. No longer store `postgres.key` in /var/vmdb/ since `postgres` can not access those files.
5. No longer configure postgres with a root cert that verifies the cert of the connecting client. This is typically `ssl_ca_file=root.crt`. We only use server side certificates and not client side certificates.

### DIFF
--- a/appliance_hardening_guide/_topics/db_ssl.md
+++ b/appliance_hardening_guide/_topics/db_ssl.md
@@ -10,14 +10,12 @@ The appliance image ships with a default SSL certificate and it is recommended t
 
 For more information, see [Generating SSL Certificates for Your Appliance and Database](#generating-ssl-certificates-for-your-appliance-and-database).
 
-
   - `/var/www/miq/vmdb/certs/server.cer` - Signed or self-signed certificate for the database appliance.
-
   - `/var/www/miq/vmdb/certs/server.cer.key` - Private key for server certificate.
 
-  - `/var/www/miq/vmdb/certs/root.crt` - The root CA certificate used to sign the CA certificate for the {{ site.data.product.title_short }} database. You can either use a self-signed certificate or a certificate that is signed by a trusted CA to generate your root certificate.
-
 It is also recommended to stop all {{ site.data.product.title_short }} services before configuring the database to use SSL.
+
+#### Configure SSL on the database appliance
 
 To configure SSL on the database appliance:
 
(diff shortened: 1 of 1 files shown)


appliance_hardening_guide/_topics/db_ssl.md
```diff
@@ -12,10 +12,11 @@ For more information, see [Generating SSL Certificates for Your Appliance and Da
 
 
   - `/var/www/miq/vmdb/certs/server.cer` - Signed or self-signed certificate for the database appliance.
-
   - `/var/www/miq/vmdb/certs/server.cer.key` - Private key for server certificate.
 
+<!--
```
Fryguy, line 17: This PR should just delete this unless it's a WIP PR.

Fryguy: Similar for the other comments

appliance_hardening_guide/_topics/db_ssl.md
```diff
@@ -83,21 +91,22 @@ The database appliance now accepts only connections from connecting appliances t
 
 1.  Log in as `root` to the connecting appliance.
 
-2.  Create the `.postgresql` directory in your `root` user home directory.
+3.  Copy the root certificate file from the database to the operating system's list of valid certificates:
 
-        [root@{{ site.data.product.title_short_l }}2 ~]# mkdir /root/.postgresql
+        [root@{{ site.data.product.title_short_l }}2 ~]# DATABASE=[database_appliance_fqdn]
```
Fryguy, line 96: I think this can be simplfied to just show a terminal prompt and doesn't need the product info...

```suggestion
        $ DATABASE=[database_appliance_fqdn]
```

Fryguy: Unless this is meant to contrast between 2 different terminal sessions?  If so, then `root@appliance-host` is just simpler

appliance_hardening_guide/_topics/db_ssl.md
```diff
@@ -83,21 +91,22 @@ The database appliance now accepts only connections from connecting appliances t
...
+        [root@{{ site.data.product.title_short_l }}2 ~]# scp root@$DATABASE:/var/www/miq/vmdb/certs/root.crt /etc/pki/ca-trust/source/$DATABASE-postgres
```
bdunne, line 97: I think this needs to be in `/etc/pki/ca-trust/source/anchors/`

appliance_hardening_guide/_topics/db_ssl.md
```diff
@@ -83,21 +91,22 @@ The database appliance now accepts only connections from connecting appliances t
...
 
-        [root@{{ site.data.product.title_short_l }}2 ~]# scp root@[database_appliance_fqdn]:/var/www/miq/vmdb/certs/root.crt /root/.postgresql/root.crt
-
-    Where `[database_appliance_fqdn]` is the fully qualified domain name of the database appliance.
+        [root@{{ site.data.product.title_short_l }}2 ~]# openssl x509 -in /etc/pki/tls/certs/ca-bundle.crt -text -noout
 
 4.  Test the connection between the connecting appliance and the database appliance by using the `psql`command:
 
-        [root@localhost ~]# psql -h [database_appliance_fqdn] -d vmdb_production
+        [root@localhost ~]# PGSSLMODE=verify-full psql -h [database_appliance_fqdn] -d vmdb_production "dbname=vmdb_production sslmode=verify-full"
```
bdunne, line 109: Is the quoted text supposed to be there?

appliance_hardening_guide/_topics/db_ssl.md
```diff
@@ -106,6 +115,9 @@ The database appliance now accepts only connections from connecting appliances t
         vmdb_production=#
 
     The `psql` displays information about the SSL connection, which indicates that the configuration succeeded. Enter `\q` to leave `psql`.
+    To understand the values for `PGSSLMODE` please see [postgress sslmode documentation].
+
+[postgres ssl documentation](https://www.postgresql.org/docs/9.0/libpq-connect.html#LIBPQ-CONNECT-SSLMODE)
```
bdunne, line 120: Should this use the v12 link? https://www.postgresql.org/docs/12/libpq-connect.html#LIBPQ-CONNECT-SSLMODE

appliance_hardening_guide/_topics/db_ssl.md
kbrock: thnx. that is how most suggested it but I put in the `ENV` var to get better success in my environment

jrafanie:
LGTM once the comments are addressed.  👍

appliance_hardening_guide/_topics/db_ssl.md
kbrock: do you want me to make this change across the board?

kbrock:
push: fixed updates

- changed unix prompt in examples
- removed comments remaining for root.crt
- fixed cert target directory to include `anchors`
- fixed postgres documentation link. Also changed the wording a little for that link.
- fixed markdown for postgres documentation link
- added headings for each section
- added check/command to ensure

Fryguy (approved):

appliance_hardening_guide/_topics/db_ssl.md
```diff
@@ -106,6 +93,17 @@ The database appliance now accepts only connections from connecting appliances t
         vmdb_production=#
 
     The `psql` displays information about the SSL connection, which indicates that the configuration succeeded. Enter `\q` to leave `psql`.
+    The `PGSSLMODE` variable forces a certain level of SSL. To learn more, please see [postgress sslmode documentation].
+
+    [postgress sslmode documentation]: https://www.postgresql.org/docs/14/libpq-connect.html#LIBPQ-CONNECT-SSLMODE
+
+5.  Ensure `/etc/default/manageiq.properties` has the proper ssl key file location defined:
+
+        $ `grep PGSSLROOTCERT /etc/default/manageiq.properties ` || \
```
Fryguy, line 102: Super minor, but these backticks within a code block are just weird to me.  Minor, but can we do:

```suggestion
        $ $(grep PGSSLROOTCERT /etc/default/manageiq.properties) || \
```

appliance_hardening_guide/_topics/db_ssl.md
```diff
@@ -106,6 +93,17 @@ The database appliance now accepts only connections from connecting appliances t
         vmdb_production=#
 
     The `psql` displays information about the SSL connection, which indicates that the configuration succeeded. Enter `\q` to leave `psql`.
+    The `PGSSLMODE` variable forces a certain level of SSL. To learn more, please see [postgress sslmode documentation].
+
+    [postgress sslmode documentation]: https://www.postgresql.org/docs/14/libpq-connect.html#LIBPQ-CONNECT-SSLMODE
```
Fryguy, line 98: If we're gonna use 14, then might as well use current, so we don't have to keep updating.

```suggestion
    [postgress sslmode documentation]: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLMODE
```

appliance_hardening_guide/_topics/db_ssl.md
```diff
@@ -74,30 +58,33 @@ To configure SSL on the database appliance:
...
-        [root@{{ site.data.product.title_short_l }}2 ~]# scp root@[database_appliance_fqdn]:/var/www/miq/vmdb/certs/root.crt /root/.postgresql/root.crt
+3.  Test the certificate got added to the certificate database:
 
-    Where `[database_appliance_fqdn]` is the fully qualified domain name of the database appliance.
+        $ openssl x509 -in /etc/pki/tls/certs/ca-bundle.crt -text -noout
 
 4.  Test the connection between the connecting appliance and the database appliance by using the `psql`command:
 
-        [root@localhost ~]# psql -h [database_appliance_fqdn] -d vmdb_production
+        $ PGSSLMODE=verify-full psql -h [database_appliance_fqdn] -d vmdb_production
```
Fryguy, line 87: I wonder if it's simpler, instead of psql followed by a separate doc line to do \q, to just have a single line something like:

```sh
$ PGSSLMODE=verify-full psql -h [database_appliance_fqdn] -d vmdb_production -c "SELECT * FROM pg_stat_ssl"
```

or perhaps filtered by only the current pid if there is more than one connection:

```sh
$ PGSSLMODE=verify-full psql -h [database_appliance_fqdn] -d vmdb_production -c "SELECT * from pg_stat_ssl WHERE pid=pg_backend_pid()"
```

Fryguy (approved):
LGTM - @bdunne Can you take another look?

appliance_hardening_guide/_topics/db_ssl.md
kbrock: hmm. translate what I typed

> thanks. made that change.

PR #1621, opened 2021-12-08
Add jekyll render directives to nested pages

agrare:


Fryguy:
@agrare Almost there - seems we need the left-side nav updated as well. e.g. https://github.com/ManageIQ/manageiq-documentation/blob/master/_data/site_menu.yml#L79

PR #1623, opened 2021-12-09
Move providers under type sections in left-size nav

agrare:
Nest e.g. infra_managers under `Administration/Managing Providers/Infrastructure Providers`
rather than them all being under one big section


agrare:
@Fryguy I know you mentioned moving managing providers around under the Administration section, we could tackle that here too

PR #1626, opened 2021-12-22
Update Appliance install on VMWare to remove configure DB steps.

dehawkins512:
See downstream issue
https://github.ibm.com/ibmprivatecloud/cp4mcm/issues/20947

### DIFF
--- a/_includes/configuration.md
+++ b/_includes/configuration.md
@@ -1,152 +1,47 @@
 ## Configuring {{ site.data.product.title_short }}
 
-After installing {{ site.data.product.title_short }} and running it for the first
-time, you must perform some basic configuration. To configure
-{{ site.data.product.title_short }}, you must at a minimum:
+After you install {{ site.data.product.title_short }} and run it for the first time, you must perform some basic configuration. To configure {{ site.data.product.title_short }}, you must complete these steps:
 
-1.  Add a disk to the infrastructure hosting your appliance.
+- Add a disk to the infrastructure that is hosting your appliance.
 
-2.  Configure the database.
-
-Configure the {{ site.data.product.title_short }} appliance using the internal
-appliance console.
+Configure the {{ site.data.product.title_short }} appliance by using the internal appliance console.
 
 ### Accessing the Appliance Console
 
-1.  Start the appliance and open a terminal console.
+1. Start the appliance and open a terminal console.
 
-2.  Enter the `appliance_console` command. The {{ site.data.product.title_short }} appliance
-    summary screen displays.
+2. Enter the `appliance_console` command. The {{ site.data.product.title_short }} appliance summary screen displays.
 
-3.  Press `Enter` to manually configure settings.
+3. Press `Enter` to manually configure settings.
 
-4.  Press the number for the item you want to change, and press `Enter`.
-    The options for your selection are displayed.
+4. Press the number for the item you want to change, and press `Enter`. The options for your selection are displayed.
 
-5.  Follow the prompts to make the changes.
+5. Follow the prompts to make the changes.
 
-6.  Press `Enter` to accept a setting where applicable.
+6. Press `Enter` to accept a setting where applicable.
 
 **Note:**
 
-The {{ site.data.product.title_short }} appliance console automatically logs out
-after five minutes of inactivity.
-
-### Configuring a Database
-
-{{ site.data.product.title_short }} uses a database to store information about the
-environment. Before using {{ site.data.product.title_short }}, configure the database
-options for it; {{ site.data.product.title_short }} provides the following two
-options for database configuration:
-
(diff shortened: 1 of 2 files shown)


dehawkins512:
@Fryguy This PR removes the steps to configure a database after appliance install.
Do we want to remove the steps for both internal and external from the install / config steps?

_includes/configuration.md
```diff
@@ -1,152 +1,47 @@
...
-5.  Follow the prompts to make the changes.
+5. Follow the prompts to make the changes.
 
-6.  Press `Enter` to accept a setting where applicable.
+6. Press `Enter` to accept a setting where applicable.
 
 **Note:**
 
-The {{ site.data.product.title_short }} appliance console automatically logs out
-after five minutes of inactivity.
```
Fryguy, line 33: I think this note should stay, because it's still relevant.

_includes/configuration.md
```diff
@@ -1,152 +1,47 @@
...
-
-  - Install an internal PostgreSQL database to the appliance
-
-  - Configure the appliance to use an external PostgreSQL database
-
-#### Configuring an Internal Database
-
-{% include configuration-db.md %}
-
-#### Configuring an External Database
```
Fryguy, line 50: This should stay...we still support external databases.  It's only the internal database that's technically wrong because we come with one out of the box, but the instructions say otherwise.

_includes/configuration.md
```diff
@@ -1,152 +1,47 @@
...
-multiple zones, such as database zone, user interface zone, and
-reporting zone, where each zone provides a specific function. The
-appliances in these zones must be configured to use an external
-database.
-
-The `postgresql.conf` file used with {{ site.data.product.title_short }} databases requires
-specific settings for correct operation. For example, it must correctly
-reclaim table space, control session timeouts, and format the PostgreSQL
-server log for improved system support. Due to these requirements, Red
-Hat recommends that external {{ site.data.product.title_short }} databases use a
```
Fryguy, line 64: When keeping this section, we can drop the Red Hat bit.  I would just say "it is recommended that"

_includes/configuration.md
```diff
@@ -1,152 +1,47 @@
...
-Ensure you configure the settings in the `postgresql.conf` to suit your
-system. For example, customize the `shared_buffers` setting according to
-the amount of real storage available in the external system hosting the
-PostgreSQL instance. In addition, depending on the aggregate number of
-appliances expected to connect to the PostgreSQL instance, it may be
-necessary to alter the `max_connections` setting.
-
-**Note:**
-
-  - {{ site.data.product.title_short }} requires PostgreSQL version 9.5.
```
Fryguy, line 77: This is actually wrong and should be PG 10.

_includes/configuration.md
```diff
@@ -1,152 +1,47 @@
...
-    summary screen displays.
-
-3.  Press **Enter** to manually configure settings.
-
-4.  Select **Configure Application** from the menu.
-
-5.  You are prompted to create or fetch a security key.
-
-      - If this is the first {{ site.data.product.title_short }} appliance, choose **Create
-        key**.
```
Fryguy, line 96: This line should probably be removed.  By default a key should have already been generated by this point.

Fryguy: Actually this whole number 5 is confusing...I don't know why we'd need this at all if the key already exists...I'll have to check on a real appliance.

_includes/configuration.md
Fryguy: Oh sorry - the diff didn't show it here, but I see it down below https://github.com/ManageIQ/manageiq-documentation/pull/1626/files#diff-719df78f71b606f44ff55a693b088d12bba38925fbdb81427af7ec7d0a9f7c43R25

_includes/configuration.md
```diff
@@ -1,152 +1,47 @@
...
-options for it; {{ site.data.product.title_short }} provides the following two
-options for database configuration:
-
-  - Install an internal PostgreSQL database to the appliance
-
-  - Configure the appliance to use an external PostgreSQL database
-
-#### Configuring an Internal Database
-
-{% include configuration-db.md %}
```
Fryguy, line 48: Should this file be deleted as well, or is it used elsewhere?

Fryguy:
@dehawkins512 Can you also remove my handle from the commit message? (I'll get pinged forever - cc @jrafanie 🤣 )

_includes/configuration.md
dehawkins512: The configuration-db.md file is also used in the quickstart openstack provider guide.

dehawkins512:
Note: These changes were merged in PR  https://github.com/ManageIQ/manageiq-documentation/pull/1641

PR #1627, opened 2021-12-22
Remove incorrect note about disabled providers

Fryguy:
@agrare Please review.

~~Alternatively, should we remove this section or reword it?  It's meant for downstream cases where the permissions file is present, but we don't have one upstream.~~ EDIT: Removing the section entirely, because it's wrong...see discussion

### DIFF
--- a/managing_providers/cloud_providers/_topics/disabling_amazon_cloud_regions.md
+++ b/managing_providers/cloud_providers/_topics/disabling_amazon_cloud_regions.md
@@ -25,9 +25,3 @@ available when adding an Amazon EC2 provider.
           - ap-northeast-1
 
 6.  Click **Save**.
-
-**Note:**
-
-In AWS, Government regions are disabled by default. To enable a disabled
-region, be sure to do so in the `production.yml` configuration file
-manually.


agrare:
Yeah if this isn't an issue upstream seems like it should be removed and only exist in the downstream docs cc @dehawkins512

Fryguy:
We really should have _something_ in the upstream docs for the permissions file, but perhaps not make it AWS specific?

agrare:
> We really should have something in the upstream docs for the permissions file, but perhaps not make it AWS specific?

Maybe?  We don't have a permissions.yml upstream at all though right?

Fryguy:
Right, but if a user creates a permissions.yml for their installation, it will be honored.

Fryguy:
It's kind of weird, because it's specifically meant for downstream distributions that don't even want to allow enabling via settings.  It's like an uber-disable.

agrare:
> Right, but if a user creates a permissions.yml for their installation, it will be honored.

Yeah good point, definitely not AWS specific but maybe in the top-level "Managing Providers" overview?

Fryguy:
> Yeah good point, definitely not AWS specific but maybe in the top-level "Managing Providers" overview?

Yeah that's probably a good place.  Generically, any provider could be added to the permissions file to prevent it from even being honored.

Actually now that I wrote that, this line is completely wrong (or maybe I'm misinterpreting it?)  The permissions.yml prevents the _provider_ from showing up, not specific regions.  Specific regions are handled with the normal Advanced Settings.

agrare:
I think this must have been from when this was a whitelist not a blacklist?

Actually wait, regions aren't in permissions.yml only top-level providers are

Fryguy:
Ok, so here's what I think we should do.  We should still add a section to Managing Providers about permissions.yml.  Separately, we should just remove this line entirely...I'll update this PR to do the latter.

Fryguy:
Ok, updated with a complete removal.

agrare (approved):

PR #1632, opened 2022-01-14
Selecting a VMware Remote Console Type not required

agrare:
The selection of a remote_console_type in Settings is no longer required as of https://github.com/ManageIQ/manageiq-ui-classic/pull/5761


Fryguy:
Should this go back to lasker? (Since that PR was in lasker)

PR #1642, opened 2022-02-07
added minimal PowerVC version requirement

iv1111:


iv1111:
@agrare hey, since the code PR is merged can we get this one merged too so that users would know which versions are supported

agrare (approved):

PR #1654, opened 2022-03-16
Drop usage of site.data.product.title_abbr

Fryguy:
The only usage of this prodcutization string is the the azure install,
and it's only used in command line examples.  The actual values don't
require it, and generic strings can be used instead.  Additionally, it's
possible to productize with values that can create invalid command line
strings.  As such, this commit removes it entirely in favor of generic
strings.

@agrare Please review.
cc @rwellon

### DIFF
--- a/installing_on_microsoft_azure/_topics/installation.md
+++ b/installing_on_microsoft_azure/_topics/installation.md
@@ -351,7 +351,7 @@ using the Azure Command-Line Interface (Azure CLI).
     minutes. Note: Enter `az storage container list` to get the list of
     storage containers.
 
-        $ az storage blob upload --account-name <storage-account-name> --container-name <container-name> --type page --file <path-to-vhd> --name <image-name>.vhd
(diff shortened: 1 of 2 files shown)


installing_on_microsoft_azure/_topics/installation.md
```diff
@@ -351,7 +351,7 @@ using the Azure Command-Line Interface (Azure CLI).
     minutes. Note: Enter `az storage container list` to get the list of
     storage containers.
 
-        $ az storage blob upload --account-name <storage-account-name> --container-name <container-name> --type page --file <path-to-vhd> --name <image-name>.vhd
+        $ az storage blob upload --account-name <storage-account-name> --container-name <storage-container-name> --type page --file <path-to-vhd> --name <image-name>.vhd
```
Fryguy, line 354: This change is just for backreference to the change below so it makes more sense.

rwellon:
This looks good to me.

agrare (approved):

PR #1678, opened 2022-06-10
Update rhv capacity setup for 4.3 and 4.4

agrare:
RHV 4.4 does not use postgres from software collections so the config is the same as 4.0 and 4.1.  I do not have a RHV 4.3 to test with so that is unknown at this point I will try to find out.

NOTE I also changed the username from `cfme` to `manageiq`

### DIFF
--- a/deployment_planning_guide/_topics/capacity.md
+++ b/deployment_planning_guide/_topics/capacity.md
@@ -66,20 +66,20 @@ database (ovirt\_engine\_history) is located. Usually, this is the RHV-M server.
 
 3. Create the user for {{ site.data.product.title_short }} to be granted read-only access to the history database:
 
-        $ psql -U postgres -c "CREATE ROLE cfme WITH LOGIN ENCRYPTED PASSWORD '[password]';" -d ovirt_engine_history
+        $ psql -U postgres -c "CREATE ROLE example_role_name WITH LOGIN ENCRYPTED PASSWORD '[password]';" -d ovirt_engine_history
 
 4. Grant the newly created user permission to connect to the history
     database:
 
-        $ psql -U postgres -c "GRANT CONNECT ON DATABASE ovirt_engine_history TO cfme;"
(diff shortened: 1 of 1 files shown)


agrare:
According to https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.3/html/data_warehouse_guide/allowing_read_only_access_to_the_history_database it appears RHV 4.3 also uses SCL postgres, I will update the docs accordingly

deployment_planning_guide/_topics/capacity.md
```diff
@@ -58,28 +58,28 @@ database (ovirt\_engine\_history) is located. Usually, this is the RHV-M server.
...
+    For RHV 4.2 and 4.3, the PostgreSQL database is delivered as a software collection in version 9.5 and must be enabled first. Therefore, to run the following psql commands, you need to enable the *rh-postgresql95* collection and load into the current shell prompt by using the source command:
 
     $ source /opt/rh/rh-postgresql95/enable
 
     </div>
 
 3. Create the user for {{ site.data.product.title_short }} to be granted read-only access to the history database:
 
-        $ psql -U postgres -c "CREATE ROLE cfme WITH LOGIN ENCRYPTED PASSWORD '[password]';" -d ovirt_engine_history
+        $ psql -U postgres -c "CREATE ROLE manageiq WITH LOGIN ENCRYPTED PASSWORD '[password]';" -d ovirt_engine_history
```
Fryguy, line 69: This role name can be anything, right?  wondering if we should productize this.

Fryguy: Or something random like "example_role_name"

agrare: yeah true, would `title_short_l` work downstream as well?

agrare: Alright update to `example_role_name`

PR #1692, opened 2022-10-05
Appliances no longer initialize on first boot

jrafanie:
We now always assume the user has attached an unpartitioned database disk and we have to configure the application in the appliance_console.

As of the following PRs, we do not initialize the application on first boot: 

https://github.com/ManageIQ/manageiq-appliance-build/pull/511
https://github.com/ManageIQ/manageiq-appliance/pull/359


agrare (approved):

jrafanie:
Oparin is the first release we didn't automatically initialize the db, so I have requested only an oparin documentation backport.

PR #1699, opened 2022-10-26
Add oracle cloud provisioning support to capabilities matrix

agrare:
https://github.com/ManageIQ/manageiq-providers-oracle_cloud/pull/79


agrare:
@Fryguy can you take a look at this?

PR #1701, opened 2022-10-28
Fix image import

jaywcarman:
Currently these links are broken in both upstream and downstream documentation. Originally we thought it was a problem with the relative link, but digging deeper I noticed that downstream requires the 'html' extension while upstream works both with and without the extension.

### DIFF
--- a/managing_providers/_topics/image_import.md
+++ b/managing_providers/_topics/image_import.md
@@ -129,7 +129,7 @@ The  following commands are to be run on a PowerVC server that is dedicated to s
 
 ### Add a Cloud Object Storage provider in {{ site.data.product.title_short }}: 
 
-1. See the corresponding instructions [here](storage_providers/ibm_cloud_object_storage_managers).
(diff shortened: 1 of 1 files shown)


Fryguy:
This shouldn't be necessary - as part of the build all .md file links are converted to .html anyway.

Fryguy:
That being said, there are other instances of references to html file, so I'll merge anyway, and we can worry about the .md -> .html conversion separately.

Fryguy:
Backported to `oparin` in commit 52ef42dbf03d5f700553c13f3a5b8a74bf4c1602.

```text
commit 52ef42dbf03d5f700553c13f3a5b8a74bf4c1602
Author: Jason Frey <fryguy9@gmail.com>
Date:   Mon Oct 31 15:59:38 2022 -0400

    Merge pull request #1701 from jaywcarman/fix_image_import
    
    Fix image import
    
    (cherry picked from commit a7a5ab583a7daba781c032716624af3722df3157)
```

PR #1706, opened 2022-12-14
Add a document on installing the VMware VDDK on containers

bdunne:
Based on:
- https://github.com/ManageIQ/manageiq-pods/issues/538
- https://github.com/ManageIQ/manageiq-pods/pull/666
- https://github.com/ManageIQ/manageiq-pods/issues/635

Fixes: https://github.com/ManageIQ/manageiq-documentation/issues/1479

### DIFF
--- a/installing_on_kubernetes/_topics/installing_vddk.md
+++ b/installing_on_kubernetes/_topics/installing_vddk.md
@@ -1,0 +1,64 @@
+## Installing the VMware-vix-disklib (VDDK) on containers
+
+To install the VDDK on containers we can take an existing container and create another layer on top of it.  Once built, this image will need to be pushed to a registry and the CR updated to reflect that.
+
+1. Fetch the VDDK linux tar file from VMware. (`VMware-vix-disklib-[version].x86_64.tar.gz`)
+    **Note:**
+
+      - If you do not already have a login ID to VMware, then you will
+        need to create one. At the time of this writing, the file can be
+        found by navigating to menu:Downloads\[vSphere\]. Select the
+        version from the drop-down list, then click the **Drivers &
+        Tools** tab. Expand **Automation Tools and SDKs**, and click
+        **Go to Downloads** next to the **VMware vSphere Virtual Disk
+        Development Kit** version. Alternatively, find the file by
+        searching for it using the **Search** on the VMware site.
+
+      - See VMware documentation for information about their policy
+        concerning backward and forward compatibility for VDDK.
+
+2. Set up a temporary directory where we will build the image
+```bash
+$ mkdir -p /tmp/vddk_container/container-assets
+```
+
+3. Copy the tar file into the container-assets directory.  For Example:
+```bash
+$ cp ~/Downloads/VMware-vix-disklib-*.tar.gz /tmp/vddk_container/container-assets
+```
+
+4. Add the following Dockerfile as `/tmp/vddk_container/Dockerfile` and substitute the second `FROM` based on the image and version that you're running.
+```dockerfile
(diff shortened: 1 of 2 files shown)


installing_on_kubernetes/_topics/installing_vddk.md
```diff
@@ -0,0 +1,64 @@
...
+        concerning backward and forward compatibility for VDDK.
+
+2. Set up a temporary directory where we will build the image
+```bash
+$ mkdir -p /tmp/vddk_container/container-assets
+```
+
+3. Copy the tar file into the container-assets directory.  For Example:
+```bash
+$ cp ~/Downloads/VMware-vix-disklib-7.0.1-16860560.x86_64.tar.gz /tmp/vddk_container/container-assets
```
jrafanie, line 27: Should this be a generic version value and we point to another document describing what versions of vddk we support and against which versions of vsphere?  cc @agrare

installing_on_kubernetes/_topics/installing_vddk.md
```diff
@@ -0,0 +1,64 @@
...
+RUN cd /vddk && \
+    filename=$(find . -name VMware-*) && \
+    if [[ -f $filename ]] ; then tar -zxvf VMware-* ; fi && \
+    mkdir -p vmware-vix-disklib-distrib && \
+    touch /vddk/vmware-vix-disklib-distrib/.keep
+
+################################################################################
+
+### IMPORTANT: Modify the following image and tag as necessary to reflect the version that you're running
+FROM docker.io/manageiq/manageiq-base-worker:latest
```
jrafanie, line 45: Should this  point to a non-existing image and tag so they can't forget to update this value?

installing_on_kubernetes/_topics/installing_vddk.md
```diff
@@ -0,0 +1,64 @@
...
+COPY --from=vddk /vddk/vmware-vix-disklib-distrib/ /usr/lib/vmware-vix-disklib/
+
+RUN /tmp/install-vmware-vddk
+```
+
+5. Build the image and tag it appropriately for your registry, then push it to the registry.
+```bash
+$ cd /tmp/vddk_container/
+$ docker build . -t your_registry.example.com/namespace/manageiq-base-worker:latest_vddk
+$ docker push your_registry.example.com/namespace/manageiq-base-worker:latest_vddk
```
jrafanie, line 56: I like this example registry, image and tag.  They can't forget to update it and the it's very clear what needs to be done here.

installing_on_kubernetes/_topics/installing_vddk.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+6. Update the CR to use the base worker image you just pushed to the registry.
+```yaml
+baseWorkerImage: your_registry.example.com/namespace/manageiq-base-worker:latest_vddk
+```
+
+The operator will now update the orchestrator and worker deployments to reflect this change.
```
jrafanie, line 64: What do they need to do if we release a new image?  Ideally, it's just changing the 
`FROM ...` and updating the revision of their image tag they build and push, right?

bdunne: Yes, they will need to rebuild anytime we update the image.

installing_on_kubernetes/_topics/installing_vddk.md
bdunne: Done

installing_on_kubernetes/_topics/installing_vddk.md
bdunne: Done

jrafanie (approved):
LGTM

PR #1710, opened 2023-01-09
Update miq 14/15 repo links

tjyang:
Update miq 14/15 repo links


Fryguy:
@tjyang We are looking at a way to remove this list to avoid having to maintain it.  See #1712.

PR #1714, opened 2023-01-11
Drop OCPv3 from the documentation

agrare:
Drop the obvious OCPv3 mentions in the docs.  There are some other sections which have to be reviewed for accuracy as I suspect they are not accurate for OCPv4 (mainly https://github.com/ManageIQ/manageiq-documentation/blob/c663fae30c18c506ca70e957bbb4d55036ba714a/integration_with_openshift_container_platform/_topics/ocp-metrics.md and the various https://github.com/ManageIQ/manageiq-documentation/blob/c663fae30c18c506ca70e957bbb4d55036ba714a/_includes/provider-ocp-mgt-token.md )

### DIFF
--- a/capabilities_matrix/_topics/container_providers.md
+++ b/capabilities_matrix/_topics/container_providers.md
@@ -2,27 +2,27 @@
 
 The following table outlines the capabilities of {{ site.data.product.title_short }} for various container providers.
 
-| Feature                                 | OpenShift 3.11 | OpenShift 4.x | Kubernetes | Amazon EKS           | Google GKE | Azure AKS | Oracle OKE | IBM Cloud IKS | VMware Tanzu |
-| --------------------------------------- | -------------- |-------------- | ---------- | -------------------- | ---------- | --------- | ---------- | ------------- | ------------ |
-| Relationship Discovery                  | ✅             | ✅            | ✅         | ✅                   |✅          |✅         |✅          |✅             |✅             |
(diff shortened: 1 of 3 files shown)


capabilities_matrix/_topics/container_providers.md
```diff
@@ -2,27 +2,27 @@
...
-| Smart State                             | ✅             | ❌            | ❌         | ❌                   |❌          |❌         |❌          |❌             |❌             |
-| OpenSCAP Execution and Report           | ✅             | ❌            | ✅         | ❌                   |❌          |❌         |❌          |❌             |❌             |
-| Cross Linking Underlying Infrastructure | ✅             | ✅            | ✅         | ✅                   |✅          |✅         |✅          |✅             |✅             |
-| Tag Mapping from Provider               | ✅             | ✅            | ✅         | ✅                   |✅          |✅         |✅          |✅             |✅             |
-| Tag Mapping to Provider                 | ❌             | ❌            | ❌         | ❌                   |❌          |❌         |❌          |❌             |❌             |
+| Feature                                 |  OpenShift 4.x | Kubernetes | Amazon EKS           | Google GKE | Azure AKS | Oracle OKE | IBM Cloud IKS | VMware Tanzu |
```
Fryguy, line 5: You can _probably_ remove the 4.x bit as well since we don't put versions for others.  If we want to set a minimum we can probably do it earlier in the doc when we list out the providers.  Up to you.

agrare: :+1: yeah good point, I'll do this when I follow-up with the changes for the token / metrics setup

PR #1728, opened 2023-03-31
Add documentation for Kafka keystores

nasark:
- added documentation outlining steps to create keystores for Kafka and how to integrate them with `internal-certificates-secret` in order to use SSL with Kafka

Depends on:
- [ ] https://github.com/ManageIQ/manageiq-pods/pull/955

Needed for:
- https://github.com/ManageIQ/manageiq-pods/pull/951

Ref:
- https://github.com/ManageIQ/manageiq-pods/issues/789

@miq-bot assign @bdunne
@miq-bot add_reviewer @Fryguy
@miq-bot add_label enhancement

### DIFF
--- a/installing_on_kubernetes/_topics/pods_ssl.md
+++ b/installing_on_kubernetes/_topics/pods_ssl.md
@@ -8,8 +8,9 @@ The certificates should all be signed by a CA and that CA certificate should be 
       --from-file=root_crt=./certs/root.crt \
       --from-file=httpd_crt=./certs/httpd.crt \
       --from-file=httpd_key=./certs/httpd.key \
-      --from-file=kafka_crt=./certs/kafka.crt \
-      --from-file=kafka_key=./certs/kafka.key \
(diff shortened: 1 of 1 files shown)


agrare:
@bdunne how is this done for other services like postgres/memcached?  I don't think there are extra manual steps for those?

nasark:
@agrare they use the cert-generator script https://github.com/ManageIQ/manageiq-pods/blob/master/tools/cert_generator.rb. Kafka uses keystores for SSL which require the use of `keytool` and so I provided manual steps. We could add a script that does this instead perhaps?

Kafka does support `PEM` format but the keys generated by the cert-generator script are in `PKSC#1` whereas Kafka uses `PKSC#8` and I was not able to make the conversion

nasark:
- removed manual steps docs and added info regarding keystore generating script added in https://github.com/ManageIQ/manageiq-pods/pull/955

nasark:
This is no longer applicable as we have moved to using strimzi

PR #1741, opened 2023-09-01
Add root key to internal certificate secret example

nasark:
- the root key is needed for downstream Kafka SSL configuration, see https://strimzi.io/docs/operators/in-development/deploying.html#installing-your-own-ca-certificates-str

@miq-bot assign @bdunne 
@miq-bot add_reviewer @Fryguy 
@miq-bot add_label enhancement

### DIFF
--- a/installing_on_kubernetes/_topics/pods_ssl.md
+++ b/installing_on_kubernetes/_topics/pods_ssl.md
@@ -6,6 +6,7 @@ The certificates should all be signed by a CA and that CA certificate should be 
 ```sh
     oc create secret generic internal-certificates-secret \
       --from-file=root_crt=./certs/root.crt \
+      --from-file=root_key=./certs/root.key \
       --from-file=httpd_crt=./certs/httpd.crt \
       --from-file=httpd_key=./certs/httpd.key \
       --from-file=kafka_crt=./certs/kafka.crt \


bdunne:
> the root key is needed for downstream Kafka SSL configuration

Why would any server need the CA private key?
It shouldn't even need the CA public key unless it is attempting to connect to itself (for debugging purposes) or to another server signed by that CA

nasark:
> Why would any server need the CA private key? This feels like a CVE waiting to happen
> It shouldn't even need the CA public key unless it is attempting to connect to itself (for debugging purposes) or to another server signed by that CA

@bdunne It's not required by us but rather Strimzi does a check to see if the following secrets are available https://strimzi.io/docs/operators/in-development/deploying.html#installing-your-own-ca-certificates-str. If `<cluster_name>-cluster-ca` secret which contains the ca key is not available then Kafka is not deployed. Usually Strimzi generates the ca key/certs and creates the secrets for you but since we are bringing our own certs in the form of `internalCertificateSecret` then it needs to be specified here

Fryguy:
Backported to `quinteros` in commit af0da36deb8457ed0bc9adb839e504d0538db220.

```text
commit af0da36deb8457ed0bc9adb839e504d0538db220
Author: Jason Frey <fryguy9@gmail.com>
Date:   Fri Sep 1 09:35:30 2023 -0400

    Merge pull request #1741 from nasark/add_root_key_pods_ssl
    
    Add root key to internal certificate secret example
    
    (cherry picked from commit 3d2f26d15affc13255d4af94ac4fe55e6167a372)
```

PR #1754, opened 2023-10-19
pg_log changed to log in postgres 10 as default log directory

jrafanie:
https://www.postgresql.org/docs/9.6/runtime-config-logging.html

log_directory (string)
When logging_collector is enabled, this parameter determines the directory in which log files will be created. It can be specified as an absolute path, or relative to the cluster data directory. This parameter can only be set in the postgresql.conf file or on the server command line. The default is pg_log.

vs.

https://www.postgresql.org/docs/10/runtime-config-logging.html#RUNTIME-CONFIG-LOGGING-WHERE log_directory (string)
When logging_collector is enabled, this parameter determines the directory in which log files will be created. It can be specified as an absolute path, or relative to the cluster data directory. This parameter can only be set in the postgresql.conf file or on the server command line. The default is log.

It remains the same in postgresql 13.

See also:
https://github.com/ManageIQ/manageiq/pull/22747
https://github.com/ManageIQ/manageiq-appliance/pull/377

### DIFF
--- a/_includes/example-postgresql-config-file.md
+++ b/_includes/example-postgresql-config-file.md
@@ -295,7 +295,7 @@
 
     # These are only used if logging_collector is on:
     log_directory = '/www/postgres/log' # Customer specific setting
-    #log_directory = 'pg_log'       # directory where log files are written,
+    #log_directory = 'log'       # directory where log files are written,
                         # can be absolute or relative to PGDATA
     log_filename = 'postgresql-%Y-%m-%d.log'        # log file name pattern,
                         # can include strftime() escapes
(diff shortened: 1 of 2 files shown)


bdunne (approved):

Fryguy:
Backported to `quinteros` in commit 61556557059f243a464fec59aa98d0c9a4a7aff5.

```text
commit 61556557059f243a464fec59aa98d0c9a4a7aff5
Author: Brandon Dunne <brandondunne@hotmail.com>
Date:   Mon Oct 23 16:14:04 2023 -0400

    Merge pull request #1754 from jrafanie/pg_log_is_now_log
    
    pg_log changed to log in postgres 10 as default log directory
    
    (cherry picked from commit 5881a8f4f1b5b966a3fbac1a2dc2446a4a65900a)
```

PR #1767, opened 2023-11-27
It is not required to stop services before backup

jrafanie:
pg_basebackup docs even describe taking online backups and even concurrent backups. Perhaps this was a requirement at one time but is no longer needed.

https://www.postgresql.org/docs/13/app-pgbasebackup.html

https://www.postgresql.org/docs/13/app-pgdump.html

### DIFF
--- a/appliances/backup_and_restore.md
+++ b/appliances/backup_and_restore.md
@@ -5,9 +5,8 @@
 Creating a backup of the appliance will be done using SSH or the console.
 
 1. Launch `appliance_console`
-   1. Select `Stop EVM Server Processes` to stop the server processes
-   2. Select `Create Database Backup` and choose a backup location (/tmp/evm_db.backup is the default)
-   3. Select `Quit` to return to the shell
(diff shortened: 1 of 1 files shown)


jrafanie:
> I thought we were using pg_dump, not pg_basebackup. If it's the former, does that change your original statement?

We recommend both.  They're both supported and have different use cases. We're actually advising here to use pg_basebackup for one time backups and pg_dump for scheduled backups a bit further down.

Neither one requires the services be shut down to run a backup/dump.

Fryguy:
Backported to `quinteros` in commit d69191b2f78cf577285c8a97192ccaa19145eaa5.

```text
commit d69191b2f78cf577285c8a97192ccaa19145eaa5
Author: Brandon Dunne <brandondunne@hotmail.com>
Date:   Mon Nov 27 11:21:46 2023 -0500

    Merge pull request #1767 from jrafanie/remove_stop_evm_server_processes_backup_prereq
    
    It is not required to stop services before backup
    
    (cherry picked from commit 67aba48f1ece59df3308ec0a6932888aa2f817ec)
```

PR #1772, opened 2023-12-13
Messaging required when upgrading to quinteros

nasark:
Since Kafka will be required, users that are upgrading will need to configure messaging via appliance_console

@miq-bot assign @agrare 
@miq-bot add_reviewer @Fryguy 
@miq-bot add_label enhancement, quinteros/yes?

### DIFF
--- a/appliances/upgrading.md
+++ b/appliances/upgrading.md
@@ -31,6 +31,8 @@ In this example, we will update from jansa to kasparov.
         # reboot
 
 
+**Note:** If upgrading to Quinteros, messaging (Kafka) setup is required. See [Configure Messaging](../_includes/configuration.md#configure-messaging)
+
 # Appendix
 
 ## RPM URLs for manageiq-release


appliances/upgrading.md
```diff
@@ -31,6 +31,8 @@ In this example, we will update from jansa to kasparov.
         # reboot
 
 
+**Note:** If upgrading to Quinteros, messaging (Kafka) setup is required. See [here](https://github.com/ManageIQ/manageiq-documentation/blob/master/_includes/configuration.md#configure-messaging)
```
nasark, line 34: Not sure if this is the correct way to link to another doc, let me know

Fryguy: No, you wouldn't link to the includes this way.  It's hard to tell but first you need to include the partial in a full doc.  Then you would like to the full doc via an absolute path from the root.

I can help you out over voice if that helps.

nasark:
After having a discussion its been decided that this is not necessary because `Configure Messaging` steps are already included in the install instructions

PR #1774, opened 2024-01-12
QA Create encrypted-parameters.md

qahmed1998:


encrypted-parameters.md
```diff
@@ -0,0 +1,50 @@
...
+Input Parameters
+
+|                           |                           |                              |
+| ------------------------- | ------------------------- |----------------------------- |
+| **Input Name**            | **Data Type**             | **Default value**            |
+| this                      | string                    | that                         |
+| manageiq_validate_certs   | boolean                   | false                        |
+| scrambled_this            | password                  | ********                     |
+
+{: caption="Table. Input Paramaters" caption-side="bottom"}
```
Fryguy, line 21: This will not work on upstream docs and should be removed.

encrypted-parameters.md
```diff
@@ -0,0 +1,50 @@
...
+## Encrypted Input Parameters
+
+An input parameter can be defined as being of type “password”, for example:
+
+Input Parameters
+
+|                           |                           |                              |
+| ------------------------- | ------------------------- |----------------------------- |
+| **Input Name**            | **Data Type**             | **Default value**            |
+| this                      | string                    | that                         |
```
Fryguy, line 17:
```suggestion
| **Input Name**            | **Data Type**             | **Default value**            |
| ------------------------- | ------------------------- |----------------------------- |
| this                      | string                    | that                         |
```

Fryguy: The formatting looks strange, but the line before the `| ------------------------- |` is the header, so the header lines should live there.  Note that you don't need to bold it because renderers will do that automatically.

encrypted-parameters.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+
+See the following table for the illustration of input parameters:
+
+Input Parameters
+
+|                           |                           |                                                                                                                                |
+| ------------------------- | ------------------------- |------------------------------------------------------------------------------------------------------------------------------- |
+| **Input Name**            | **Data Type**             | **Default value**                                                                                                              |
```
Fryguy, line 45:
```suggestion
| Input Name              | Data Type                 | Default value                                                                                                                  |
| ------------------------- | ------------------------- |------------------------------------------------------------------------------------------------------------------------------- |
```

encrypted-parameters.md
```diff
@@ -0,0 +1,50 @@
...
+| manageiq_validate_certs   | boolean                   | false                                                                                                                          |
+| cores_per_socket          | string                    | ${/#miq_provision.miq_provision_request.get_option(:dialog).fetch(dialog_option_0_cores_per_socket)}                           |
+| root_password             | string                    | password::${/#miq_provision.miq_provision_request.get_option(:dialog).fetch(password::dialog_option_0_root_password)}          |
+
+{: caption="Table. Input Paramaters" caption-side="bottom"}
```
Fryguy, line 50: Also remove this one

Fryguy:
@qahmed1998 How will this be presented?  Will it be a standalone page or included in the API docs or in Automate docs?  I ask because it's presented like an automated doc, so hard to tell.

Fryguy:
As discussed, we just have to figure out where this will live and then move it / include it accordingly.  Content-wise this looks great!

PR #1783, opened 2024-03-11
Radjabov appliance upgrade

bdunne:

### DIFF
--- a/appliances/backup_and_restore.md
+++ b/appliances/backup_and_restore.md
@@ -43,7 +43,7 @@ Restoring a backup of the appliance will be done using SSH or the console.
      rm -f /backup.tgz # if desired to save space
    ```
 3. Launch `appliance_console`
-   1. If this is a brand new appliance and the database has not yet been initialized, select `Configure Application`. If this is an existing database appliance, this step can be skipped.
+   1. Since the database has not yet been initialized, select `Configure Application`.
       1. When prompted to configure the database, select `Create Internal Database`
       2. When prompted to configure messaging, either connect to an external messaging system or configure this appliance as a messaging server
       3. Choose a database disk
--- a/appliances/backup_and_restore.md
+++ b/appliances/backup_and_restore.md
@@ -45,7 +45,7 @@ Restoring a backup of the appliance will be done using SSH or the console.
 3. Launch `appliance_console`
    1. If this is a brand new appliance and the database has not yet been initialized, select `Configure Application`. If this is an existing database appliance, this step can be skipped.
       1. When prompted to configure the database, select `Create Internal Database`
-      2. When prompted to configure messaging, select `Make No messaging changes`
+      2. When prompted to configure messaging, either connect to an external messaging system or configure this appliance as a messaging server
       3. Choose a database disk
       4. Answer y or n to "Should this appliance run as a standalone database server?"
       5. Enter a region number (0), this is not important and will be overwritten on restore
--- a/appliances/backup_and_restore.md
+++ b/appliances/backup_and_restore.md
@@ -39,8 +39,8 @@ Restoring a backup of the appliance will be done using SSH or the console.
 2. On the command line:
    ```bash
      cd /
-     tar -zxf /backup
-     rm -f /backup # if desired to save space
+     tar -zxf /backup.tgz
+     rm -f /backup.tgz # if desired to save space
    ```
 3. Launch `appliance_console`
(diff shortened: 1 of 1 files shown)


appliances/backup_and_restore.md
```diff
@@ -35,17 +35,22 @@ Before adding the cron job it would be best practice to mount an external storag
 # Restore
 Restoring a backup of the appliance will be done using SSH or the console.
 
-1. Download a backup TGZ and place it in `/`
-2. On the command line:
+1. Set the hostname on the new appliance.  On the command line:
+   ```bash
+     hostnamectl hostname your-new-hostname.example.com
+     echo "your.ip.address your-new-hostname.example.com your-new-hostname" > /etc/hosts
```
agrare, line 41: Probably should be `>> /etc/hosts` so it doesn't wipe out the `127.0.0.1 localhost, etc...`

agrare: We could probably add a command to get their IP automatically, `ip -4 -br addr show eth0 | awk '{print $3}'` just need a little regex to drop the `/subnet`

appliances/backup_and_restore.md
```diff
@@ -35,17 +35,22 @@ Before adding the cron job it would be best practice to mount an external storag
...
+     tar -zxf /backup.tgz
+     rm -f /backup.tgz # if desired to save space
    ```
-3. Launch `appliance_console`
-   1. If this is a brand new appliance and the database has not yet been initialized, select `Configure Application`. If this is an existing database appliance, this step can be skipped.
+4. Launch `appliance_console`
+   1. Since the database has not yet been initialized, select `Configure Application`.
       1. When prompted to configure the database, select `Create Internal Database`
-      2. When prompted to configure messaging, select `Make No messaging changes`
+      2. When prompted to configure messaging, either connect to an external messaging system or configure this appliance as a messaging server
```
jrafanie, line 53: Should we say something such as:

If you are planning on restoring a database to this system, please verify the encryption key, v2_key, for use with that database was properly extracted from the backup.tgz into the certs directory.  Ideally, we could add a prompt in the console where it displays the "generate new key" screen to also prompt that an existing v2_key was found in the else case.  I'm not sure how we streamline this other than to prompt for the backup.tgz and we extract it and look for the v2_key and warn if it's missing.

appliances/backup_and_restore.md
```diff
@@ -35,17 +35,22 @@ Before adding the cron job it would be best practice to mount an external storag
...
+3. On the command line:
    ```bash
      cd /
-     tar -zxf /backup
-     rm -f /backup # if desired to save space
+     tar -zxf /backup.tgz
+     rm -f /backup.tgz # if desired to save space
    ```
-3. Launch `appliance_console`
-   1. If this is a brand new appliance and the database has not yet been initialized, select `Configure Application`. If this is an existing database appliance, this step can be skipped.
```
mikemoroney, line 46: There are 2 possible scenarios 
1. restoring from a backup to the same appliance  - in this case, there is no need to create a database 
2. restoring from a backup to a new appliance - in this case, it is necessary to create a database as per the steps below

jrafanie: 1) True.  The problem is configuring the database on a brand new appliance does more than create a database.  It takes the empty disk, sets it up with physical and logical volumes and initializes postgres including the data directory but also the configuration and permissions.  I probably missed other things it's doing.  So, we can't really restore a database directly on a "new" appliance before all of these things are done.  That's why this section is here.  If we can move to asking for the tgz file, from the backup step, we can detect if there's no v2_key for the backup and warn the user.  That way, we can ensure the v2_key is copied over before we try to initialize postgresql and do the restore.

I think for now, we have to do the simplest things we can do to make this more clear.

jrafanie:
@bdunne what are your thoughts on the comments?  I'm fine with this but always think we should clarify the need for the v2_key to be extracted before you try to restore the database.

mikemoroney (approved):
Thanks - LGTM

jrafanie:
I'm fine with merging as is and we can add more if needed.

Fryguy:
Backported to `radjabov` in commit f59f573b1796434fc9b485e75f4c1947cd95115e.

```text
commit f59f573b1796434fc9b485e75f4c1947cd95115e
Author: Joe Rafaniello <jrafanie@users.noreply.github.com>
Date:   Mon Mar 25 14:09:31 2024 -0400

    Merge pull request #1783 from bdunne/radjabov_appliance_upgrade
    
    Radjabov appliance upgrade
    
    (cherry picked from commit 665ed7455d5d24cab2e3f67ced60b443087b9ebd)
```

PR #1806, opened 2024-08-22
Updated ToC for entire ManageIQ + added already existing topics within ToC

qahmed1998:
Added following additional topics to ToC:
- Creating a Service in for Virtual Machine Provisioning
- Service Provisioning Using an Orchestration Template
- Assigning a Custom Analysis Profile to a Virtual Machine

This PR is editing the current ToC for ManageIQ. No files in terms of their locations were moved in the process. 3 existing topics (as mentioned above) were added to the ToC.

PR #1812, opened 2024-09-11
Fix indenting issue in site_menu

Fryguy:
Fixes a bug introduced in #1806 where the indenting was causing 

```
Psych::SyntaxError: (/tmp/manageiq-documentation/_data/site_menu.yml): mapping values are not allowed in this context at line 7 column 17
```

I don't understand why that PR didn't fail considering it does a full build, but I have a feeling it's not using the PR for the build, but instead using master.

@jrafanie Please review.

### DIFF
--- a/_data/site_menu.yml
+++ b/_data/site_menu.yml
@@ -4,36 +4,36 @@ installing:
   title: Installing
   children:
     - title: Containerized (podified)
-        children:
-           - title: Kubernetes
-             path: "/docs/reference/latest/installing_on_kubernetes/index.html"
-             desc: "How to install and configure ManageIQ on Kubernetes"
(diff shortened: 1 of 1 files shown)


jrafanie:
> I don't understand why that PR didn't fail considering it does a full build, but I have a feeling it's not using the PR for the build, but instead using master.

😮

Fryguy:
LOL, so this went red and I think it's exactly for the same reason the last PR went green.  I think the build is pulling from master instead of the PR.  I'm not sure how to fix this offhand, so I'm ok with merging this on red to get master back to green.  @jrafanie it's up to you

jrafanie (approved):

PR #1823, opened 2024-12-04
Add instructions on enabling VMware Request Debugging

agrare:
https://github.com/ManageIQ/manageiq-providers-vmware/pull/930

### DIFF
--- a/troubleshooting/_topics/providers.md
+++ b/troubleshooting/_topics/providers.md
@@ -1,0 +1,51 @@
+## Providers
+
+### VMware vCenter Provider
+
+If you are having issues with a VMware provider you can enable extensive debug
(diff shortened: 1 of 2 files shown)


troubleshooting/_topics/providers.md
agrare, line 1: I wasn't sure if it was better to put this in troubleshooting or in the managing_providers/infrastructure_providers/vmware_vcenter_providers but since it really shouldn't be necessary in the normal course of operation thought troubleshooting was the right place.

Fryguy: Yeah I think troubleshooting is better.

Fryguy (approved):

PR #1845, opened 2025-04-29
Fix broken links on API services page

Fryguy:
@kbrock Please review.


api/reference/services.md
```diff
@@ -702,18 +699,15 @@ Or via the delete action for deleting multiple services:
 }
 ```
```
Fryguy, line 700: This is "broken" in github highlighting because the json above here has a `...` which is invalid json.  However, it renders correctly in reality.

kbrock (approved):

PR #1849, opened 2025-06-09
Add caveat about KubeVirt VMs from Instance Types

agrare:
Currently only KubeVirt VMs created from Templates work with VM Reconfigure.

Related: https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/296


capabilities_matrix/_topics/infrastructure_providers.md
```diff
@@ -64,6 +64,8 @@
 | Create Network Router                                        | ❌      | ✅          | ❌                      | ❌            | ❌          |
 | Delete Network Router                                        | ❌      | ✅          | ❌                      | ❌            | ❌          |
 
+\* KubeVirt/OSV VMs created with an instance type are not supported for reconfigure
```
agrare, line 67: NOTE `\` here because I want this to look like an asterisk not a bullet point

PR #1850, opened 2025-06-11
Delete accidentally added automation_providers directory

agrare:
These were part of a WIP refactoring of the managing_providers/_topics/automation_management_providers section but were accidentally added in an unrelated change in 18b02fe90b49883bf6590be06cd17686c4f831f1

Fixes https://github.com/ManageIQ/manageiq.org/actions/runs/15547403500/job/43771461115#step:4:107


Fryguy (approved):

Fryguy:
This is failing because docs build always pulls from master and doesn't consider the PR :(

PR #1851, opened 2025-06-19
Add a guide for migrating from pods to appliances

bdunne:
TODO:
- [x] run through another restore pass on a clean appliance

### DIFF
--- a/installing_on_kubernetes/_topics/migrate_to_appliance.md
+++ b/installing_on_kubernetes/_topics/migrate_to_appliance.md
@@ -20,7 +20,7 @@
    # Copy the database backup to the appliance (usually under /tmp/ or somewhere that has enough space)
    ```
 
-1. Extraxt the database password, encryption key, server GUID and region number
+1. Extract the database password, encryption key, server GUID and region number
    ```bash
    $ oc rsh deploy/orchestrator
    sh-5.1$ cd /var/www/miq/vmdb/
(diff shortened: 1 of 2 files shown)


installing_on_kubernetes/_topics/migrate_to_appliance.md
```diff
@@ -0,0 +1,102 @@
...
+   $ psql -d vmdb_production -c "SELECT pg_size_pretty(pg_database_size('vmdb_production'));"
+   $ pg_dump -Fc -d vmdb_production > /var/lib/pgsql/data/postgres.dump
+   # Copy the database backup to the appliance (usually under /tmp/ or somewhere that has enough space)
+   ```
+
+1. Extraxt the database password, encryption key, server GUID and region number
+   ```bash
+   $ oc rsh deploy/orchestrator
+   sh-5.1$ cd /var/www/miq/vmdb/
+   sh-5.1$ source ./container_env
```
jrafanie, line 27: Do we need to `source ./container_env` if we're just cat'ing files below?  Maybe you loaded rails console previously and forgot to remove this step.

installing_on_kubernetes/_topics/migrate_to_appliance.md
```diff
@@ -0,0 +1,102 @@
+## Migrating from Kubernetes to Appliances
+
+### Prerequisites
+1. Deploy an appliance with a database disk.  It is recommended that the appliance being deployed is the same code version as the kubernetes environment.
```
jrafanie, line 4: Do we want to provide instructions to support going to a newer version?

I feel like it's safer to migrate to the same version and then do an immediate upgrade via rpm.  What do you think?

jrafanie: For followup

jrafanie (approved):
LGTM, I pushed some typo fixes and think the other comments can done in followups

PR #1864, opened 2025-09-09
Update openshift management token instructions

agrare:
Service Account Tokens are no longer automatically generated when the service account gets created, `oc sa new-token` is deprecated, and `oc create token` has a limited duration (even if it can be years) and has not well documented default.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-long-lived-api-token-for-a-serviceaccount describes how to create a long-lived service account token which matches what `oc sa new-token` would generate.

### DIFF
--- a/_includes/provider-ocp-mgt-token.md
+++ b/_includes/provider-ocp-mgt-token.md
@@ -1,18 +1,33 @@
 Run the following to obtain the token needed to add a Red Hat OpenShift provider:
 
-1.  Obtain the `management` service account token name:
+1. Obtain the service account token name:
+    ```bash
+    oc apply -f - <<EOF
+    apiVersion: v1
+    kind: Secret
+    metadata:
+     name: $service_account_name-secret
+     namespace: $project_name
+     annotations:
+       kubernetes.io/service-account.name: $service_account_name
+    type: kubernetes.io/service-account-token
+    EOF
 
-        # oc describe sa -n $project_name $service_account_name
-        ...
-        Tokens:  management-admin-token-0f3fh
-                 management-admin-token-q7a87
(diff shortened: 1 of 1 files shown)


_includes/provider-ocp-mgt-token.md
```diff
@@ -1,18 +1,33 @@
...
+     annotations:
+       kubernetes.io/service-account.name: $service_account_name
+    type: kubernetes.io/service-account-token
+    EOF
 
-        # oc describe sa -n $project_name $service_account_name
-        ...
-        Tokens:  management-admin-token-0f3fh
-                 management-admin-token-q7a87
+    oc describe secret -n $project_name $service_account_name-secret
```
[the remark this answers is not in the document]
Fryguy: Sorry, yes `oc get`

_includes/provider-ocp-mgt-token.md
```diff
@@ -1,18 +1,21 @@
...
-    output, replacing `management-admin-token-0f3fh` with the name of
-    your token:
-
-        # oc describe secret -n $project_name management-admin-token-0f3fh
-        ...
-        Data
-        ====
-        token:  eyJhbGciOiJSUzI1NiI...
+   ```bash
+   oc get secret -o template $service_account_name-secret -n $project_name --template={{.data.token}
```
Fryguy, line 19: think there's a typo in the template. Also, I think it's clearer if the template lives next to the -o template, so:


```suggestion
   oc get secret -n $project_name -o template --template={{.data.token}} $service_account_name-secret
```

agrare: Oh yeah missed last character while copy&pasting, fixed

PR #1893, opened 2026-06-11
Add instructions for upgrading podified to tal & PG16

bdunne:


appliances/upgrading.md
```diff
@@ -35,3 +35,63 @@ In this example, we will update from jansa to kasparov.
...
+1. Switch to the namespace where the application is installed.
+
+        # oc project <your_namespace>
+
+2. Upgrade the operator (and image references other than postgresqlVersion)
+
+        # oc edit deployment {{ title_short_l }}-operator
+          # Change .spec.template.spec.containers[0].image to the new image
+        # oc edit {{ site.data.product.operator_custom_resource_definition_name_plural }}
+          # Update all of the image tags except for postgresql
```
Fryguy: Instead of commenting everything I think we should use proper backticks with `bash` language. Then the comments can be used for the prose parts. (Same with the other sections).

When using backticks, be sure to indent 3 (or 4) spaces so it sits indented under the numbered bullet instead of flush left.

Example:

````md
1. Hi

```bash
echo "Hi"
```
````

renders as:

1. Hi

```bash
echo "Hi"
```

---
as opposed to indenting the backticks:

---

````md
1. Hi

   ```bash
   echo "Hi"
   ```
````

renders as:

1. Hi

   ```bash
   echo "Hi"
   ```

Comment (jrafanie) [reply appliances/upgrading.md]: > # Update all of the image tags except for postgresql

this feels error prone... I wonder if we should enumerate them or capture them beforehand so we can verify they're all updated afterwards.

File: appliances/upgrading.md [hunk 2]
```diff
@@ -35,3 +35,63 @@ In this example, we will update from jansa to kasparov.
...
+          # Change the .spec.postgresqlImage tag from "13-el9" to "16-el10"
+
+7. The new postgresql pod will get deployed, but stuck in CrashLoopBackOff because of incompatible data directory
+
+        # oc delete pvc/postgresql deployment/postgresql
+
+8. Copy the database backup to the new pvc
+
+        # oc get pod -l name=postgresql
+        # oc cp vmdb.pg.gz <name of new postgresql-xxx pod>:/var/lib/pgsql/data/vmdb.pg.gz
```
jrafanie: Will the prior step prevent it from remaining in a CrackLoopBackOff? Does the operator recreate the pg deployment and pvc empty?  Does the pod start properly without an initialized db?

bdunne: Yes, deleting the deployment and pvc will cause the operator to deploy a brand new database.

appliances/upgrading.md
jrafanie: sorry, I didn't see this was resolved

jrafanie (approved):
LGTM, sorry, I didn't realize the fixes were pushed.

Fryguy:
Backported to `tal` in commit 8c268886779e1b5c3bd4a945266b581238f09d26.

```text
commit 8c268886779e1b5c3bd4a945266b581238f09d26
Author: Joe Rafaniello <jrafanie@users.noreply.github.com>
Date:   Fri Jul 24 09:25:20 2026 -0400

    Merge pull request #1893 from bdunne/pods_upgrade_to_pg16
    
    Add instructions for upgrading podified to tal & PG16
    
    (cherry picked from commit 7ca6c748f4b6156fbb95842d0b1777401ae97acc)
```

PR #1900, opened 2026-06-29
Add OIDCPassClaimsAs default value to OIDC configuration docs

jrafanie:
mod_auth_openidc 2.4.15+ encodes claim values as ISO-8859-1 when passing them in headers/env vars. Characters outside the Latin-1 range cause an invalid byte sequence error. Setting OIDCPassClaimsAs to none disables this encoding for backwards compatibility.

Related: https://github.com/ManageIQ/manageiq-pods/pull/1406
https://github.com/ManageIQ/manageiq-appliance/pull/409

### DIFF
--- a/auth/openid_connect.md
+++ b/auth/openid_connect.md
@@ -141,6 +141,7 @@ The defaults in the *manageiq-external-auth-oidc.conf* file are:
 | LogLevel                 | warn      |
 | OIDCCryptoPassphrase     | sp-cookie |
 | OIDCOAuthRemoteUserClaim | username  |
+| OIDCPassClaimsAs         | both none |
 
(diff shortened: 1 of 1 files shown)


Fryguy (commented):
Not sure I understand this change as it's not installation specific? That is, when would a user ever change this?

jrafanie:
> Not sure I understand this change as it's not installation specific? That is, when would a user ever change this?

You're correct, it's certainly not a required change and likely not a user facing setting: https://github.com/OpenIDC/mod_auth_openidc/blob/a72ef34be4ae4e90114ffab5ec06095dd6c11f0e/auth_openidc.conf#L928-L943, closing.

